PR #20, opened 2016-01-15
Batch reduce

andyfaff:
Adding @llimeht batch reduction code


andyfaff:
@llimeht I merged the batch_reduce branch into master, not sure that it was the most elegant way of doing it, but still...

PR #21, opened 2016-01-22
ENH: adding runtests.py, build and test without installing

andyfaff:
ENH: adding runtests.py, build and test without installing


andyfaff:
@llimeht, build and test without install.

PR #29, opened 2016-04-28
Fixed parallelisation

andyfaff:


andyfaff:
The bug in the parallelisation did not affect the correctness of the reflectivity calculation

PR #31, opened 2016-04-28
ENH speedup use  np.convolve vs fftconvolve

andyfaff:


andyfaff:
Improves constant dq/q smearing by ~20%

PR #32, opened 2016-05-06
Added NIST tests to CurveFitter

andyfaff:


andyfaff:
This PR adds the NIST tests for non-linear regression to refnx. CurveFitter (and by extension lmfit and scipy) is tested against all those examples. One of those NIST tests is a 2D fit, which wasn't previously being tested.

PR #63, opened 2017-04-25
ENH:Manual beam finder

andyfaff:
Create a manual beam finder for locating a specular beam


andyfaff:
Still to do: draggable foreground/background regions.

PR #85, opened 2017-08-11
ENH: reversed structure

andyfaff:
add `reversed` attribute to `refnx.reflect.Structure`. This enables you to calculate the front/back reflectivity from the same structure very easily.

### DIFF
--- a/refnx/reflect/structure.py
+++ b/refnx/reflect/structure.py
@@ -17,7 +17,36 @@ class Structure(UserList):
     Represents the interfacial Structure of a reflectometry sample. Successive
     Components are added to the Structure to construct the interface.
     """
-    def __init__(self, name='', solvent='backing'):
+    def __init__(self, name='', solvent='backing', reversed=False):
+        """
+        Represents the interfacial Structure of a reflectometry sample.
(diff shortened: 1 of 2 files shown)


andyfaff:
@igresh - comments?

igresh:
I can see no problems with the implementation. This change will make using refnx for a series of related fits in a notebook much cleaner.

There is an argument to be made that this feature should be  accessed through the model rather than the structure (as its technically not a structural change, but a change to the experimental setup) - but if this should happen it can happen later on.

andyfaff:
Can you think of a better name than reversed?

igresh:
If it's accessed through structure reversed makes good sense. 

If it's accessed through model then something like Neutron_Incidence = "Back" or "Bottom" / "Front" or "Top" would make sense (you loose the Boolean, but it's more readable IMO).

andyfaff:
I did it through Structure because Structure also calculates the sld
profile.

igresh:
It's current implementation makes sense (and I'm happy with it) but it could be implemented in any class with access to the slabs (I think).

PR #95, opened 2017-09-12
Fix reflectometry analysis

arm61:
MAINT: Small fix for reflectometry_analysis Jupyter Notebook
Removal of the nbins option from the CurveFitter.sample optionals has been accounted for in the notebook now, using syntax outlined in the docstrings

n_gen has changed to ngen in the objective.pgen function, this change is now accounted for in the example notebook.


arm61:
Closed due to duplication.

PR #97, opened 2017-10-14
BUG: ensure fname_dat set in all code paths

llimeht:
If `reduce_stitch` is used with `save=False` then `fname_dat` is not defined leading to a `NameError`.


andyfaff:
I'd rather `fname_dat` was set before the conditional, rather than adding an `else`. I've just made a change to do that.

PR #98, opened 2017-10-14
BUG: fix run number selection from cache

llimeht:
Code to extract runs based on run numbers was collecting only the indexes in the cache not the actual data.


andyfaff:
Does a test need to be added for this?

PR #100, opened 2017-10-14
Make bg subtraction configurable per angle

llimeht:
(probably an interim change to the API to be eventually replaced by some sort of ReducerOptions configuration)

### DIFF
--- a/refnx/reduce/reduce.py
+++ b/refnx/reduce/reduce.py
@@ -463,8 +463,9 @@ class PlatypusReduce(object):
         return deepcopy(reduction)
 
 
-def reduce_stitch(reflect_list, direct_list, norm_file_num=None,
-                  data_folder=None, trim_trailing=True, save=True, **kwds):
+def reduce_stitch(reflect_list, direct_list, background_list=None,
+                  norm_file_num=None, data_folder=None, trim_trailing=True,
+                  save=True, **kwds):
     """
     Reduces a list of reflected beam run numbers and a list of corresponding
     direct beam run numbers from the Platypus reflectometer. If there are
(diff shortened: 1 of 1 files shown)


refnx/reduce/reduce.py
```diff
@@ -463,8 +463,9 @@ def _reduce_single_angle(self, scale=1):
         return deepcopy(reduction)
 
 
-def reduce_stitch(reflect_list, direct_list, norm_file_num=None,
-                  data_folder=None, trim_trailing=True, save=True, **kwds):
+def reduce_stitch(reflect_list, direct_list, background_list=None,
```
andyfaff: Would it be better here to refactor reflect_list such along the lines of:
`[(708, config_dict_for_first_angle), (709, config_dict_for_second_angle)]`? Then we could get rid of direct_list and **kwds and background_list

refnx/reduce/reduce.py
```diff
@@ -477,6 +478,9 @@ def reduce_stitch(reflect_list, direct_list, norm_file_num=None,
         708 corresponds to the file PLP0000708.nx.hdf.
     direct_list : list
         Direct beam run numbers, e.g. `[711, 711, 711]`
+    background_list : list, optional
+        List of bool to control whether background subtraction is used for each
```
andyfaff: Should `bool` be in quotes, `'bool'`

refnx/reduce/reduce.py
llimeht: Eventually, something like that would make sense. Getting what we are currently using documented and committed and then later moving to config objects feels like the right approach.

PR #101, opened 2017-10-14
Allow beam finder tolerance to be varied

llimeht:
For some runs, setting the tolerance on the auto beamfinder is required so that the finder converges as it moves across the y-t image. The default 1% convergence criterion might not be met by the right part of the image, resulting in a 30-60px beam width; relaxing convergence to 2 or 3% allows a 2-3px beam width to be determined with 2 or 3 iterations before convergence.


refnx/reduce/platypusnexus.py
```diff
@@ -722,8 +722,10 @@ def process(self, h5norm=None, lo_wavelength=2.5, hi_wavelength=19.,
 
         elif peak_pos is None:
             # use the auto finder, falling back to manual_beam_find
+            tol = kwds.get('peak_pos_tol', 0.01)
```
andyfaff: peak_pos_tol needs to be added to the keywords in the function signature, and the docstring amended.

PR #99, opened 2017-10-14
Batch reducer improvements

llimeht:
Improve documentation and usability of batch reducer, including:

* adding an on-disk cache of reduced data so that the reducer can be restarted and plots can be made without requiring all data to be rereduced.

* making the list of reduced runs prettier by setting column types to `int` where appropriate and replacing `NaN` with blank entries.

PR #102, opened 2017-10-14
Add a config file to the batch reducer

llimeht:
(Builds on PR #99 "Batch reducer improvements" so cumulative changes are shown here until merging/rebasing is performed)

This is a first experiment at adding a config file to control various reducer variables for the batch reducer. In the long run, it makes sense to push more options into a config file and a config class, allowing the API to be simplified with fewer keyword arguments required and fewer hard-to-track-down `**kwargs` being passed from layer to layer.


andyfaff:
At what point does `yaml` have to be importable? I'd prefer for it to be delayed until some function from batchreduction is used. At the moment it's imported at the top of batchreduction.py. I don't really want to make it a required dependency (for everything apart from batchreduction).

e.g.
import refnx
import refnx.analysis
import refnx.reduction

shouldn't cause yaml to be imported.

refnx/reduce/batchreduction.py
```diff
@@ -427,3 +590,52 @@ def run_list(entry, mode='refl'):
...
+            return list(conf.keys())[0]
+
+        self.default_name = _name(configs[0])
+
+        for config in configs:
+            name = _name(config)
+            self[name] = config[name]
+
+    @classmethod
+    def from_yaml_string(cls, string):
```
andyfaff: perhaps do the importing of yaml here?

refnx/reduce/batchreduction.py
```diff
@@ -427,3 +590,52 @@ def run_list(entry, mode='refl'):
...
+        """ create a BatchReducerConfiguration from a YAML string """
+        conf = yaml.safe_load(io.StringIO(string))
+        return cls(conf)
+
+    @classmethod
+    def from_yaml_file(cls, filename):
+        """ create a BatchReducerConfiguration from a YAML file """
```
andyfaff: perhaps do the importing of yaml here.

setup.py
```diff
@@ -120,7 +120,7 @@ def write_version_py(filename='refnx/version.py'):
         'include_package_data': True,
         'setup_requires': ['numpy'],
         'install_requires': ['numpy', 'scipy', 'emcee', 'six',
-                             'uncertainties', 'pandas']
+                             'uncertainties', 'pandas', 'pyyaml']
```
andyfaff: I don't want to make it a required dependency. Makes it a lot harder for people to install from source (esp. Windows) if there are issues installing from conda/pip (or there's a version problem).

llimeht:
eww? importing things in functions is horrible. A try/except around the import with a _HAS_YAML being defined would work but that's still very clumsy and delays start-up errors until runtime.

If `import refnx.reduction` would now require `yaml` then that exposes a latent bug in `__init__.py`. It probably should not be importing and fiddling with the namespace in this way; same goes for the requirement for `pandas`, btw. If the user hasn't imported the batch reduction code, they shouldn't need its dependencies.

That said, `pyyaml` is a pretty standard package and is trivially installable via conda, pip, apt etc. Given how much relies on it, it's also quite stable. 

As per our discussions last week and some of the comments in #100, the long term future of the reducer is probably to grow a configuration object and to lose the explosion of keyword arguments that get passed around. The reducer config object needs a human-editable file format for the user to set the correct parameters for the set of reductions. YAML makes good sense as a format to use as it is standardised, sufficiently expressive, widely used and low on punctuation (cf json or xml); my feeling is to go with more yaml rather than less.

refnx/reduce/batchreduction.py
```diff
@@ -251,33 +358,61 @@ class BatchReducer:
...
         """
         Create a batch reducer using metadata from a spreadsheet
 
         Parameters
         ----------
-        filename : str
+        filename : str, optional
             The filename of the spreadsheet to be used. Must be readable by
-            `pandas.read_excel` (`.xls` and `.xlsx` files).
+            `pandas.read_excel` (`.xls` and `.xlsx` files). If not specified,
```
andyfaff: Does pandas use xlrd to read Excel files? If not then I can't remember why xlrd is required as a dependency

refnx/reduce/batchreduction.py
```diff
@@ -427,3 +590,52 @@ def run_list(entry, mode='refl'):
 
     # valid = [int(r) for r in l if not np.isnan(r)]
     return [int(v) for v in valid]
+
+
+class BatchReducerConfiguration(dict):
+    """ Configuration for the batch reducer
+
+    Example:
```
andyfaff: We need an example of a configuration file, and what key:value are allowed. Otherwise I'd have no idea.

andyfaff: Also, is there a point to having a batch reduction configuration if in the future we have configuration files for each reflect:direct configuration. 
Off the top of my head the only setting one needs is whether to remove redundant points once two datasets have been spliced.

refnx/reduce/batchreduction.py
arm61: pandas does use xlrd/xlwt for reading/writing excel -- although these are both 'optional' dependencies of pandas

PR #104, opened 2017-10-15
 Batch reducer improvements

llimeht:
Improve documentation and usability of batch reducer, including:

*    adding an on-disk cache of reduced data so that the reducer can be restarted and plots can be made without requiring all data to be rereduced.

*    making the list of reduced runs prettier by setting column types to `int` where appropriate and replacing `NaN` with blank entries.

(#99 can't be reopened so recreating)


andyfaff:
If this passes I'll merge it tonight. But I haven't reviewed the code in detail.

PR #107, opened 2017-10-17
MAINT: moved process_chain to function

andyfaff:
Moved `process_chain` and `load_chain` to their own functions. This makes it simpler to load and process chains.


andyfaff:
@llimeht comments? Unfortunately this means that a lot of the example notebooks will have to be rejigged, *again*.

PR #115, opened 2017-10-31
MAINT: deal with new/old emcee w thinning

andyfaff:
- Deals with new/old emcee thinning behaviour
- Can use the new tqdm w new emcee if it's installed.


andyfaff:
Close reopen for travis

PR #133, opened 2017-11-24
ENH: add lnprob_extra function

andyfaff:
The log-prior consists of the varying parameter log-probabilities, the model log-probability (which shouldn't include any parameter log-probs).
The log-likelihood consists of the model vs data log-probability.

However, there may be an extra term which is difficult for the user to add via any of the above.
For example, the user may not be able to amend the model code. This PR adds the option of providing a `lnprob_extra` function which is folder into the log-prior term.


andyfaff:
I also seem to be doing a lot of documentation changes in this PR, which will be confusing for onlookers.

PR #137, opened 2017-11-30
Warn about missing cython at install time

llimeht:
Installation without cython is possible but should be warned against.


andyfaff:
THanks @llimeht, no point in waiting for CI to finish, all CI tests use cython.

PR #147, opened 2018-01-15
ENH: Addition of the surf_monolayer class to the analytical profiles

arm61:
This is a class for the modelling of lipid and surfactant monolayers at some interface. It offers the ability to constrain the model such that the number density of the head groups and the tail groups are consistent. Further, any number of contrasts can be compared and the model will be consistent across all contrasts (this assumes that the same surface excess is obtained between different contrasts). 

TODO: 
- Add an example notebook including neutron data (multiple contrasts).
- Allow for a mixed monolayer.
- Allow for the inclusion of a protein molecule (e.g. from PDB).


andyfaff:
Good stuff, thanks for the contribution. I'll look into it soon. Protip: if you use a notebook, add the clean notebook to git, with no cells executed. Executed notebooks lead to bloat.

arm61:
I think that it is a clean notebook that I have included (it just a long one with lots of comments). There is no rush on my side for getting it pulled (just wanted it online somewhere as it will be used in a couple of papers I am writing).

andyfaff:
I've had a closer look and I think that the `SurfMono` class is going to need to be rewritten from scratch unfortunately. For an example of a component please see the [brush.py](https://github.com/refnx/refnx/blob/master/examples/analytical_profiles/brushes/brush.py)
code.

To start with a `Component` needs to implement the `slabs`, `parameters` and `lnprob` methods/properties.
`slabs` returns the slab representation of the Component (thickness/sld/isld/roughness/vf of solvent penetration).
`parameters` should return the  `refnx.analysis.Parameters` associated with the component
`lnprob` should return any extra log-probability terms associated with the component. Note, this is not the lnprobs of the actual parameters themselves, but anything extra.

What parameters are needed to set the Component up? I'm guessing:

- head_scat_lens
- tail_scat_lens
- head_volume
- tail_volume
- head_thickness
- tail_thickness
- area_per_molecule

`sub_phase_sld, super_phase_sld` are covered by the solvent penetration. What do you do for roughness between tail/head, head/other stuff, tail/otherstuff?

There are different ways of ensuring constraints across contrasts, listed here in order of preference:

1) use the same `SurfMono` object across contrasts
2) use the same parameters to set different `SurfMono` objects up.
3) use constraints with different parameters to set different `SurfMono` objects up (this option is by far the worst.

Perhaps it's better if I draft something up?

PR #155, opened 2018-02-27
Ability to solvate the tails with air

arm61:
As mentioned in the discussion on PR #150, I have added the ability to optionally solvate the 1st layer with the 0th layer and the -2nd layer with the -1st layer by given the solvent keyword "top/bottom". As discussed I could not identify a general solution so have opted instead to share this non-general option.


andyfaff:
Functionality included in #204.

PR #156, opened 2018-02-28
ENH: solvation now done by specific solvent

andyfaff:
@arm61 @llimeht @igresh @llimeht 
This is a change that significantly alters behaviour. When a `Structure` is created one used to be able to use `fronting` or `backing` to define how you wanted to solvate the structure. 

Now the string arguments have been removed, and you have to supply an `SLD` instance.
If you don't supply an SLD instance then the SLD of the backing medium is used by default (`Structure[-1].slabs[-1]`).
This should be back compatible for those that automatically use backing medium for solvation, but those who use fronting medium for solvation need to supply an `SLD` instance.
The PR has the advantage that an arbitrary SLD can be used for solvation, it doesn't have to be the fronting or backing media.

### DIFF
--- a/examples/analytical_profiles/brushes/brush.py
+++ b/examples/analytical_profiles/brushes/brush.py
@@ -105,6 +105,12 @@ class FreeformVFPextent(Component):
         direction that is requested.
     microslab_max_thickness : float
         Thickness of microslicing of spline for reflectivity calculation.
+
+    Notes
+    -----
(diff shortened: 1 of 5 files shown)


andyfaff:
@tjmurdoch

arm61:
Would it not be more pragmatic to allow the solvent definition as part of the layer, therefore allowing the selective solvation of different layers?

PR #157, opened 2018-03-01
ENH: mixed_model

andyfaff:
Adds the ability to add create a mixed area reflectivity model.
For example, you may have a surface that is 50% covered by a monolayer, and 50% of a surface that isn't. You can create a `Structure` for each of those areas, and combine them together in a `MixedReflectModel`, with a scale factor for each of the two structures. THe scale factors are used to calculate the relative contribution of each area.

### DIFF
--- a/refnx/reflect/reflect_model.py
+++ b/refnx/reflect/reflect_model.py
@@ -631,7 +631,8 @@ class MixedReflectModel(object):
             tscales = [1 / len(structures)] * len(structures)
 
         for scale in tscales:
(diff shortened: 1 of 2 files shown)


andyfaff:
@llimeht @tjmurdoch this enables the problems we discussed yesterday afternoon.

andyfaff:
Actually, it seems that MixedReflectModel was already in, it just wasn't tested. Now it's being tested, and it's still usable for analysing mixed areas.

PR #161, opened 2018-03-19
MAINT: vendorize emcee3.

andyfaff:
This package wishes to use emcee version 3. However, it's not yet released, and the release date is unknown. The master branch is fairly stable and reasonably tested. I therefore think it's possible to have a `refnx._lib.emcee` module, which is a vendored copy of the emcee package.
When emcee version 3 is released that code can be removed, and we can go back to installing the emcee package from PyPI. If modifications are made to the emcee codebase in the meantime, then those modifications can be copied verbatim.
There is no need to fork emcee at this point.


andyfaff:
@igresh

PR #169, opened 2018-04-07
Enable in-place running of unit-tests

llimeht:
* Teach setup.py to use pytest as the runner
* Use setup.py to run the tests on travis
* Remove test runner script; setup.py and pytest can do all of this now
* Update documentation to match


andyfaff:
I have a few questions.

- I assume all test running can be done in-place (i.e. no dependence on installed version of refnx)?
- How can the tests for only one module be run (equiv. to `python runtests.py -s reflect`)?
- How can the tests for only one test-file be run (equiv. to `python runtests.py -t refnx.analysis.test:test_objective::TestObjective`)?

andyfaff:
> That neither of the commands you gave actually work for me with runtest.py

Hmm, I didn't check that they worked, but that's unusual that they didn't work. Anyway, so long as I can do the testing of the individual modules/tests, that's ok. I've just checked that that works as expected.

andyfaff:
Thanks @llimeht

PR #172, opened 2018-04-11
MAINT: check that foreground is totally within detector

andyfaff:
check that foreground is totally within detector and that background pixels are totally within detector


andyfaff:
There is also a rough check to compare the foreground width to the estimated beam size on the detector. If the two are radically different then a warning is raised. However, the test is not foolproof, but may flag more egregious cases.

PR #175, opened 2018-05-03
BUG: all Parameter instances were sharing same Interval instance

andyfaff:
All `Parameter` instances were sharing the same `Interval` instance, a reasonably bad bug.


andyfaff:
The fact that it hadn't been discovered so far meant that it probably hasn't affected user work, it's obvious when you stumble upon it.

PR #193, opened 2018-07-01
MAINT: check var pars aren't changing

andyfaff:


igresh:
This merge removes the capacity to serialise-deserialise the curvefitter object (useful for large fitting problems and sharing/saving results).

A use-case for this is:
```python
fitter.sample(10)
pickle.dump(fitter, open('fitter.pkl', 'wb'))
# Restart the session for some reason
# Decide that you want to do a bit more sampling with the fitter
fitter = pickle.load(open('fitter.pkl', 'rb'))
fitter.sample(10)
```

The issue seems to be in CurveFitter._check_vars_unchanged(), which will raise a `RuntimeError` if an identical copy (but different `id()`) of the previous `varying_parameters()` are used. 

An easy fit would be to change the error to a warning?

andyfaff:
Ok, gotcha. The PR was designed to pick up if people were chopping and changing fitting parameters during the lifetime of the curvefitter. Unfortunately there are side effects, which you've just discovered. I'll look into it.

PR #198, opened 2018-07-10
MAINT: facilitate pickling of curvefitter

andyfaff:


andyfaff:
I think the tests are passing, but there is something going on with appveyor at the moment.

PR #215, opened 2018-08-06
ENH: Scattering length density profile from a molecular dynamics simulation trajectory file.

arm61:
Addition of the ability to obtain a SLD profile directly from a molecular dynamics simulation using the `Simulation` class. This uses `MDAnalysis` to read in a pdb file and calculates the scattering length density of each layer. This currently requires a dictionary file (see `simulation.lgt`) to know the relevant scattering length for each individual atom type as defined in the simulation file.


setup.py
```diff
@@ -195,7 +195,7 @@ def setup_package():
                                 name='refnx.reduce._cevent',
                                 sources=['src/_cevent.pyx'],
                                 include_dirs=[numpy_include],
-                                language='c',
+                                language='c++',
```
arm61: This is due to a possibly local issue with my gcc compiler/cython version where the compilation would fail as a result of the -> notation that being present in the generated .c file.

andyfaff:
This will be a great addition.

refnx/reflect/simulation.py
```diff
@@ -0,0 +1,167 @@
...
+        This is to allow for the use of a vacuum gap at the bottom of the cell.
+    flip: bool, optional
+        False if the system should be read as is, true is the simulation cell
+        should be rotated through they-plane -- note that this treats the first
+        side that the neutron or X-ray interacts with as that at z=0.
+    roughness: float, optional
+        The fractional (of the layer thickness) roughness to be considered
+        between the layers.
+    verbose: bool, optional
+        True if you want to be notified when the pdb and lgt files have been
```
andyfaff: What information is provided? Does the read take a long time?

refnx/reflect/simulation.py
```diff
@@ -0,0 +1,167 @@
...
+        read and when the sld profile is calculated.
+    """
+    def __init__(self, pdbfile=None, lgtfile=None, xray=False,
+                 layer_thickness=1, cut_off=5, flip=False, roughness=0,
+                 verbose=False):
+        self.pdbfile = pdbfile
+        self.lgtfile = lgtfile
+        self.u = None
```
andyfaff: self.u needs documenting

refnx/reflect/simulation.py
```diff
@@ -0,0 +1,167 @@
+import MDAnalysis as mda
+import numpy as np
+from refnx.analysis import Parameters
+from refnx.reflect import structure
+
+
+class Simulation:
```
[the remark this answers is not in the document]
andyfaff: Should this subclass `Component` if you'd like to use it in a `Structure`?

refnx/reflect/simulation.py
```diff
@@ -0,0 +1,167 @@
...
+    def __init__(self, pdbfile=None, lgtfile=None, xray=False,
+                 layer_thickness=1, cut_off=5, flip=False, roughness=0,
+                 verbose=False):
+        self.pdbfile = pdbfile
+        self.lgtfile = lgtfile
+        self.u = None
+        self.xray = xray
+        self.scatlens = {}
+        self.read_pdb()
+        if verbose:
```
andyfaff: not sure if printing this is really necessary. Is it difficult to read a PDB file?

refnx/reflect/simulation.py
```diff
@@ -0,0 +1,167 @@
+import MDAnalysis as mda
+import numpy as np
+from refnx.analysis import Parameters
+from refnx.reflect import structure
+
+
+class Simulation:
+    """
+    Determines the scattering length density profile from a simulation output
```
andyfaff: `Determines the scattering length density profile along the z-axis of a molecular dynamics output file (PDB format). Lateral averaging occurs in the x-y plane.`

andyfaff: Is the file format generic, are there simulation cells for which this code will fail?

refnx/reflect/simulation.py
```diff
@@ -0,0 +1,167 @@
...
+        self.u = None
+        self.xray = xray
+        self.scatlens = {}
+        self.read_pdb()
+        if verbose:
+            print('PDB file read.')
+        self.read_lgt()
+        if verbose:
+            print('LGT file read.')
+        self.av_layers = np.zeros((int(self.u.dimensions[2] /
```
andyfaff: floor division?

refnx/reflect/simulation.py
```diff
@@ -0,0 +1,167 @@
...
+                                       layer_thickness) -
+                                   cut_off, 5))
+        self.layers = np.zeros((self.av_layers.shape[0],
+                                self.av_layers.shape[1],
+                                len(self.u.trajectory)))
+        self.layers[:, 0, :] = layer_thickness
+        self.layers[:, 3, :] = layer_thickness * roughness
+        self.layers[:, 4, :] = 0
```
andyfaff: not needed, array already initialised.

refnx/reflect/simulation.py
```diff
@@ -0,0 +1,167 @@
...
+        self.av_layers[:, 0] = layer_thickness
+        self.av_layers[:, 3] = layer_thickness * roughness
+        self.av_layers[:, 4] = 0
```
andyfaff: not needed, array already initialised.

refnx/reflect/simulation.py
```diff
@@ -0,0 +1,167 @@
...
+        self.av_layers = np.zeros((int(self.u.dimensions[2] /
+                                       layer_thickness) -
+                                   cut_off, 5))
+        self.layers = np.zeros((self.av_layers.shape[0],
+                                self.av_layers.shape[1],
+                                len(self.u.trajectory)))
+        self.layers[:, 0, :] = layer_thickness
+        self.layers[:, 3, :] = layer_thickness * roughness
+        self.layers[:, 4, :] = 0
+        self.av_layers[:, 0] = layer_thickness
```
andyfaff: How about assigning `self.av_layers` from `self.layers`?

refnx/reflect/simulation.py
```diff
@@ -0,0 +1,167 @@
...
+        self.get_sld_profile()
+        if verbose:
+            print('SLD profile determined.')
+
+    def read_pdb(self):
+        """Parse pdb file.
+
+        This reads the pdb file into memory.
+        """
+        self.u = mda.Universe(self.pdbfile)
```
andyfaff: Does `mda.Universe` open and close the file cleanly?

refnx/reflect/simulation.py
```diff
@@ -0,0 +1,167 @@
...
+
+    def read_lgt(self):
+        """Parses .lgt.
+
+        Parses the lgtfile. If no lgtfile is defined falass will help the user
```
andyfaff: falass? :-)

refnx/reflect/simulation.py
```diff
@@ -0,0 +1,167 @@
...
+            print('SLD profile determined.')
+
+    def read_pdb(self):
+        """Parse pdb file.
+
+        This reads the pdb file into memory.
+        """
+        self.u = mda.Universe(self.pdbfile)
+
+    def read_lgt(self):
```
andyfaff: Refactor to use `periodictable` for scattering lengths. Constructor will need to know whether X-rays or neutrons are required. Also raises the following points:

- there may be waters in the cell - can you develop a method to change the contrast on those? Then we can use the same object for multiple contrasts.
- how is isotopic substitution in general performed? Can one specify the deuteration of certain components, e.g. tail deuteration in a lipid.  It'd be really nice to have that functionality without processing the PDF file to do that. e.g. specify atom numbers whose isotope can be changed at will.

refnx/reflect/simulation.py
```diff
@@ -0,0 +1,167 @@
...
+                i = 1
+                if self.xray:
+                    i *= 2.817940
+                self.scatlens[line_list[0]] = [float(line_list[1]) * i,
+                                               float(line_list[2]) * i]
+            file.close()
+        else:
+            raise ValueError("No lgtfile has been defined.")
+
+    def get_sld_profile(self):
```
andyfaff: change to a private method, `_get_sld_profile` because it doesn't need to be accessible by the public.

refnx/reflect/simulation.py
```diff
@@ -0,0 +1,167 @@
...
+                                     self.layers[0, 0, k])
+            k += 1
+        if self.flip:
+            self.layers = self.layers[::-1, :, :]
+        self.av_layers = np.average(self.layers, axis=2)
+
+    def sld_profile(self, z=None):
+        """
+        Calculates an SLD profile, as a function of distance through the
+        interface.s
```
andyfaff: errant s.

refnx/reflect/simulation.py
```diff
@@ -0,0 +1,167 @@
...
+
+        This will calculate the SLD profile for each of the timesteps. This is
+        achieved by summing the scattering lengths for each of the atoms found
+        in a given layer (of defined thickness). The total scattering length is
+        converted to a density by division by the volume of the layer.
+        """
+        u = self.u
+        k = 0
+        for ts in u.trajectory:
+            for atom in range(0, len(u.atoms)):
```
andyfaff: does this nested loop take long?

refnx/reflect/simulation.py
```diff
@@ -0,0 +1,167 @@
...
+    def get_sld_profile(self):
+        """Calculate SLD profile.
+
+        This will calculate the SLD profile for each of the timesteps. This is
+        achieved by summing the scattering lengths for each of the atoms found
+        in a given layer (of defined thickness). The total scattering length is
+        converted to a density by division by the volume of the layer.
+        """
+        u = self.u
+        k = 0
```
andyfaff: what is k the counter for? Is it the number of trajectories? In which case: `for k, ts in enumerate(u.trajectory):`

refnx/reflect/simulation.py
```diff
@@ -0,0 +1,167 @@
...
+        for ts in u.trajectory:
+            for atom in range(0, len(u.atoms)):
+                analysis_layers = self.layers.shape[0] * self.layers[0, 0, k]
```
andyfaff: It's not immediately clear to me what the code in the loops is doing (although I can guess). Needs more comments. What are the different axes of `layers` used for? what is `analysis_layers` used for?

refnx/reflect/simulation.py
andyfaff: Can one unroll the `for atom` loop using vectorisation, etc?

refnx/reflect/test/simulation.dat
```diff
@@ -0,0 +1,51 @@
+0.051793 0.00034985 5.0354e-06
```
andyfaff: `mdsimulation.dat`

refnx/reflect/test/simulation.lgt
```diff
@@ -0,0 +1,57 @@
+C1 26.659000000000002 0.0
```
andyfaff: not needed if the scattering lengths come from `periodictable`.

refnx/reflect/test/simulation.lgt
```diff
@@ -0,0 +1,57 @@
...
+C46 19.998 0.0
+C47 19.998 0.0
+C48 19.998 0.0
+C49 19.998 0.0
+C50 19.998 0.0
+C51 19.998 0.0
+C52 19.998 0.0
+C53 19.998 0.0
+C54 26.659000000000002 0.0
+OW 0.0 0.0
```
andyfaff: water has zero scattering length... Is this meant to enable contrast variation somehow?

refnx/reflect/simulation.py
```diff
@@ -0,0 +1,167 @@
...
+                if u.atoms[atom].position[2] < analysis_layers:
+                    bin_choose = int(u.atoms[atom].position[2] /
+                                     self.layers[0, 0, k])
+                    self.layers[bin_choose, 1, k] += \
+                        self.scatlens[u.atoms[atom].name][0]
+                    self.layers[bin_choose, 2, k] += \
+                        self.scatlens[u.atoms[atom].name][1]
+            self.layers[:, 1, k] /= (u.dimensions[0] * u.dimensions[1] *
+                                     self.layers[0, 0, k])
```
andyfaff: Is the volume fraction of solvent taken into account somehow, or is there explicit solvent in the box?

setup.py
andyfaff: Interesting. Is that on macOS or linux? I wonder why this hasn't come up on travis?

refnx/reflect/simulation.py
```diff
@@ -0,0 +1,167 @@
...
+            self.layers[:, 2, k] /= (u.dimensions[0] * u.dimensions[1] *
+                                     self.layers[0, 0, k])
+            k += 1
+        if self.flip:
+            self.layers = self.layers[::-1, :, :]
+        self.av_layers = np.average(self.layers, axis=2)
```
andyfaff: what is axis=2?

andyfaff: I'm guessing time, but it's not explicitly mentioned.

setup.py
arm61: this is linux (ubuntu) with:
- cython 0.28.5
- numpy 1.15.0
- gcc 5.4.0

INSTALL_CONTRIBUTE.md
```diff
@@ -9,8 +9,7 @@ Monte Carlo to obtain posterior distributions for curvefitting problems.
 
 *refnx* has been tested on Python 2.7, 3.4, 3.5 and 3.6. It requires the *numpy,
 scipy, cython, pandas, emcee* packages to work. Additional features require the
-*pytest, h5py, xlrd, uncertainties, ptemcee, tqdm, matplotlib* packages. To build
-the bleeding edge code you will need to have access to a C-compiler to build a
+*pytest, h5py, xlrd, uncertainties, ptemcee, tqdm, matplotlib, MDAnalysis* packages. To build the bleeding edge code you will need to have access to a C-compiler to build a
```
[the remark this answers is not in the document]
arm61: I am working through these changes and will push them all in a one-er. Just going to comment "Done" on those I have dealt with.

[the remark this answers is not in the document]
arm61: Done

appveyor.yml
```diff
@@ -55,8 +55,8 @@ install:
     - cmd: call conda.exe info --envs
 
     # Configure the VM.
-    - cmd: call conda.exe install --yes --quiet numpy scipy h5py cython six traitlets ipywidgets xlrd pandas pytest
-    - pip install corner uncertainties matplotlib IPython
+    - cmd: call conda.exe install --yes --quiet numpy scipy h5py cython six traitlets ipywidgets xlrd pandas pytest MDAnalysis
```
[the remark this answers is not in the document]
arm61: Not Done. See full changes and explanation later.

refnx/reflect/__init__.py
```diff
@@ -5,6 +5,8 @@
 from refnx.reflect.structure import (Structure, SLD, Slab, Component,
                                      sld_profile)
 from refnx.reflect.spline import Spline
+from refnx.reflect.simulation import Simulation
```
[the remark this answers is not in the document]
arm61: Done.

refnx/reflect/simulation.py
```diff
@@ -0,0 +1,167 @@
+import MDAnalysis as mda
```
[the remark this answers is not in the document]
arm61: Done.

refnx/reflect/simulation.py
```diff
@@ -0,0 +1,167 @@
...
+import numpy as np
```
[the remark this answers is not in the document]
arm61: Done.

refnx/reflect/simulation.py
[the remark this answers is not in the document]
arm61: Done.

refnx/reflect/simulation.py
```diff
@@ -0,0 +1,167 @@
...
+          True if the scattering length of the particles should be scaled by
+          the classical radius of an electron.
+    layer_thickness: float
+        The thickness of the layers that the simulation cell should be sliced
+        into.
+    cut_off: float
+        The size of the simulation cell that should be ignored from the bottom.
+        This is to allow for the use of a vacuum gap at the bottom of the cell.
+    flip: bool, optional
+        False if the system should be read as is, true is the simulation cell
```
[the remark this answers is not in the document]
arm61: Done.

refnx/reflect/simulation.py
```diff
@@ -0,0 +1,167 @@
...
+        should be rotated through they-plane -- note that this treats the first
```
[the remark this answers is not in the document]
arm61: Done.

refnx/reflect/simulation.py
```diff
@@ -0,0 +1,167 @@
...
+        side that the neutron or X-ray interacts with as that at z=0.
+    roughness: float, optional
+        The fractional (of the layer thickness) roughness to be considered
```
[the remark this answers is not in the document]
arm61: Done. (made more clear)

refnx/reflect/simulation.py
arm61: Done.

refnx/reflect/simulation.py
arm61: Done. See discussion later.

refnx/reflect/simulation.py
arm61: `np.floor_division` (weirdly) produces a `float` so it is still necessary to convert to `int`, whereas `int` should always (I think) round down.

refnx/reflect/simulation.py
arm61: Done.

refnx/reflect/simulation.py
arm61: Done.

refnx/reflect/simulation.py
arm61: I believe that it does. It is quite a popular package in the MD community.

refnx/reflect/simulation.py
arm61: Haha, this is a residual from when I had my own code to do this. 
Resolved.

refnx/reflect/simulation.py
arm61: Done.

refnx/reflect/simulation.py
arm61: Done.

refnx/reflect/simulation.py
arm61: Done.

refnx/reflect/test/simulation.dat
arm61: Done. and changed all the other files appropriately.

refnx/reflect/simulation.py
```diff
@@ -0,0 +1,167 @@
...
+        where the columns are atom_type, real_scattering_length, and
+        imaginary_scattering_length respectively.
+    xray: bool, optional
+          True if the scattering length of the particles should be scaled by
+          the classical radius of an electron.
+    layer_thickness: float
+        The thickness of the layers that the simulation cell should be sliced
+        into.
+    cut_off: float
+        The size of the simulation cell that should be ignored from the bottom.
```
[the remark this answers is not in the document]
arm61: Done

refnx/reflect/simulation.py
arm61: pdb files can be very large (I have some in the region of many gigabytes) so it can be reassuring to know to some extent where you are in the process. I would have a percentage read thing similar to the emcee counter thing but using MDAnalysis limits this.

refnx/reflect/simulation.py
arm61: Done.

refnx/reflect/simulation.py
arm61: Done.

refnx/reflect/simulation.py
arm61: Done.

refnx/reflect/simulation.py
arm61: It doesn't look like the atoms in MDAnalysis are vectorisable currently

setup.py
arm61: oh and python 3.7.0

refnx/reflect/simulation.py
arm61: there is explicit solvent in the simulation (and there probably should be in most simulations). The solvent is just acmw (hence the zero scattering in the .lgts file.

refnx/reflect/test/simulation.lgt
arm61: See point 2

refnx/reflect/test/simulation.lgt
arm61: see point 2.

refnx/reflect/simulation.py
arm61: see point 2.

refnx/reflect/simulation.py
```diff
@@ -0,0 +1,167 @@
...
+    lgtfile: str
+        The path and name of the .lgt file (which contains the scattering
+        lengths of each of the atom type in the pdbfile). Currently the .lgt
+        file style that is supported is a 3 column space separated txt file
+        where the columns are atom_type, real_scattering_length, and
+        imaginary_scattering_length respectively.
+    xray: bool, optional
+          True if the scattering length of the particles should be scaled by
+          the classical radius of an electron.
+    layer_thickness: float
```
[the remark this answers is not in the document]
arm61: the value really depends on the force field that is used. for an all-atom force field I have found this to be good.

refnx/reflect/simulation.py
```diff
@@ -0,0 +1,167 @@
...
+class Simulation:
+    """
+    Determines the scattering length density profile from a simulation output
+    file.
+
+    Parameters
+    ----------
+    pdbfile: str
+        The path and name of the .pdb file
+    lgtfile: str
```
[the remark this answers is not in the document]
arm61: see point 2

refnx/reflect/simulation.py
```diff
@@ -0,0 +1,167 @@
...
+        The path and name of the .lgt file (which contains the scattering
+        lengths of each of the atom type in the pdbfile). Currently the .lgt
+        file style that is supported is a 3 column space separated txt file
+        where the columns are atom_type, real_scattering_length, and
+        imaginary_scattering_length respectively.
+    xray: bool, optional
+          True if the scattering length of the particles should be scaled by
```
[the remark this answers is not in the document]
arm61: see point 2.

refnx/reflect/simulation.py
arm61: See comment about pdb file sizes.

arm61:
Okay, I think I have managed to take care of everything!

1. Having spoken to one of the developers of MDAnalysis, it appears that it isn't yet fully windows compatible hence appveyor failing for both https://github.com/refnx/refnx/pull/215/commits/acce374743884b2ed52d798effdf8a5f0740c5e0 and https://github.com/refnx/refnx/pull/215/commits/3b13ba1152eac8d52129f3c4cf6623ea95d47227. To resolve this, I have written a simple emulation of what MDAnalysis does for use when it isn't available (the only downside is it is slower). 
2. My logic for using a custom .lgt file type was that I have been using this to look at coarse-grained forcefields (where a single bead represents many atoms), therefore something like periodictable doesn't really work. However I have now made it such that if there is no .lgt file defined the "element" (the final column in the pdb) will be used to generate the scattering length. I reckon for non-standard stuff (contrast variation things), the .lgt file is still the best option, however I am open to suggestions. For example, the system in `mdsimulation.*` represents a deuterated lipid on acmw (hence the 0. scattering for water), and the simulation used a united-atom forcefield (no hydrogens), so using the .lgt file made the most sense. 

I reckon I could do with more tests. Will work on them tomorrow (thought this might be enough to keep you busy :smile:)

andyfaff:
I hadn't thought about what one would do for a coarse grained situation. What I had in mind was that someone would simulate a membrane with MD, and compare to multiple contrast NR/XRR. The multiple contrast NR/XRR would involve one or more of the following:

- different solvent contrasts
- changing tail deuteration (one or both of the tails)
- changing head deuteration
- last two points can be generalised to one or more subsets of isotopically substituted atoms. e.g. one tail and the headgroup is deuterated.

Of course, we're using lipid membranes as an example, but it could apply to any MDsimulation. From cursory inspection of the `MDanalysis` package one can [select](https://www.mdanalysis.org/pages/basic_example/) subsets of atoms:

```
bb = u.select_atoms('protein and backbone')  # a selection (AtomGroup)
```

I was wondering if that could be used by the code you're writing to filter the atoms in the tail group and substitute the specific isotopes as required.

arm61:
Okay, I have moved the file reading to biopython (which should be more windows compatible). 

Additionally in the interest of greater contrast control, once the files have been read in it is now possible to scattering length to either specific atom type (e.g. atom type 'C1') or to specific residues (e.g. all the residues of type 'SOL' or 'DSP'). 

I feel this is a good general solution. This way, for example the tail contrast could be assigned by using the `set_atom_scattering()` function as follows

```python
sim = MDSimulation(...)

 # an array of all the hydrogen atom types present in the tails
atom_types = ['H1', 'H2', ...]
scattering_lengths = [[6.671, 0], ] * len(atom_types)
for k, type in enumerate(atom_types):
    sim.set_atom_scattering(type, scattering_length[k])
```

Similarly, if you wanted to set all of the water to be D2O, the `set_residue_scattering()` function could be used (assuming the solvent has the residue name `'SOL'`, and that the ordering was oxygen, hydrogen, hydrogen). 

```python
sim = MDSimulation(...)

sim.set_residue_scattering('SOL', [[5.803, 0], [6.671, 0], [6.671, 0]])
```

or the `set_atom_scattering()` could be used, 

```python
sim = MDSimulation(...)

sim.set_atom_scattering('HW1', [6.671, 0]) 
sim.set_atom_scattering('HW2', [6.671, 0]) 
```

This is a good general solution as the user will be familiar with how the atoms and residues are defined in their own simulation. 

Thoughts?

arm61:
Okay, that is some pretty decent testing available now. Further I have included a third way of assigning the scattering lengths where it is necessary to define two arrays; one that is all the atom types in the simulation and one that is the corresponding scattering lengths. These are then passed like so 

```python 
sim = MDSimulation(...)

sim.assign_scattering_lengths(atom_types=atom_type_array, scattering_lengths=scattering_length_array)
```

If that function is run without an arguments the element column in the pdb is used to get the scattering lengths (neutrons, for xrays it needs to be given an energy), and if a .lgt file is used this is given as an argument.

refnx/reflect/_mdsimulation.py
```diff
@@ -0,0 +1,307 @@
...
+        layers to stop overbinning.
+    cut_off: float, optional
+        The thickness (in the z-dimension) of the volume of the bottom
+        simulation cell that should be ignored. This is to allow for the use of
+        a vacuum gap in the simulation cell. This thickness can be determined
+        by visualising the simulation or by running the box with a cut-off of 0
+        and investigating the SLD profile that is produced.
+    flip: bool, optional
+        False if the system should be read as is, True if the simulation cell
+        should be rotated through the xy-plane. This expects the neutrons or
```
andyfaff: Normally one would rotate around an axis, or mirror through a plane.

refnx/reflect/_mdsimulation.py
```diff
@@ -0,0 +1,307 @@
...
+        X-ray to be incident at the bottom of the simulation cell (where
```
andyfaff: So flip means the cell is inverted and what was at the bottom of the cell is not closest to the incident beam?

refnx/reflect/_mdsimulation.py
```diff
@@ -0,0 +1,307 @@
...
+
+class MDSimulation(Component):
+    """
+    Determines the scattering length density profile along the z-axis of a
+    molecular dynamics output file (PDB format, unless MDAnalysis is
+    installed). Lateral averaging occurs in the x-y plane.
+
+    Parameters
+    ----------
+    pdbfile: str
```
andyfaff: Allow `pdbfile` to be a file-like

pdbfile: str or file-like

refnx/reflect/_mdsimulation.py
```diff
@@ -0,0 +1,307 @@
...
+        The path and name of the .pdb file or for which the reflectometry
+        should be found.
+    layer_thickness: float, optional
+        The thickness of the layers that the simulation cell should be sliced
+        into. This will depend on the size of the particles in the simulation,
+        for example a coarse-grained forcefield simulation will require thicker
+        layers to stop overbinning.
+    cut_off: float, optional
```
andyfaff: How does cut_off work? Does it cut_off before flip?

refnx/reflect/_mdsimulation.py
```diff
@@ -0,0 +1,307 @@
...
+        a vacuum gap in the simulation cell. This thickness can be determined
+        by visualising the simulation or by running the box with a cut-off of 0
+        and investigating the SLD profile that is produced.
+    flip: bool, optional
+        False if the system should be read as is, True if the simulation cell
+        should be rotated through the xy-plane. This expects the neutrons or
+        X-ray to be incident at the bottom of the simulation cell (where
+        z=0).
+    roughness: float, optional
+        The fractional (of the layer thickness) roughness to be considered
```
andyfaff: The fraction

refnx/reflect/_mdsimulation.py
```diff
@@ -0,0 +1,307 @@
...
+        between the layers in the simulation cell.
+    verbose: bool, optional
+        True if you want to be notified when the pdb and lgt files have been
```
andyfaff: Perhaps we could use a tqdm progress bar instead?

refnx/reflect/_mdsimulation.py
```diff
@@ -0,0 +1,307 @@
...
+            print('SLD profile determined.')
+
+    def read_pdb(self):
+        """Parse pdb file.
+
+        Parses the pdbfile.
+        """
+        parser = PDBParser()
+        self.structure = parser.get_structure('model', self.pdbfile)
+        file = open(self.pdbfile, 'r')
```
andyfaff: Don't use `file`, use `f`.
Also can we open the file in a context manager pov:

```
with possibly_open_file(pdbfile, 'r') as f:
   # do stuff
```

`possibly_open_file` is a context manager in `refnx._lib`

refnx/reflect/_mdsimulation.py
```diff
@@ -0,0 +1,307 @@
...
+
+    def run(self):
+        """
+        Calculate the scattering profile from the simulation trajectory.
+        """
+        self._get_sld_profile()
+        if self.verbose:
+            print('SLD profile determined.')
+
+    def read_pdb(self):
```
andyfaff: `read_pdb` should probably be a standalone function returning a `structure` and dimensions. THere's no real need for it to be an object method.

refnx/reflect/_mdsimulation.py
```diff
@@ -0,0 +1,307 @@
...
+        self.cut_off = cut_off
+
+    def assign_scattering_lengths(self, lgtfile=None, radiation='neutron',
+                                  xray_energy=None, atom_types=None,
+                                  scattering_lengths=None):
+        """
+        Assign the scattering lengths either from a .lgt file, a user defined
+        array or allow it to be determined from the elements listed in the
+        .pdb file.
+        Parameters
```
andyfaff: \n before `Parameters`

refnx/reflect/_mdsimulation.py
```diff
@@ -0,0 +1,307 @@
...
+        self.av_layers[:, 3] = layer_thickness * roughness
+        self.layers = np.array([self.av_layers, ] * len(self.structure))
+        self.flip = flip
+        self.cut_off = cut_off
+
+    def assign_scattering_lengths(self, lgtfile=None, radiation='neutron',
+                                  xray_energy=None, atom_types=None,
+                                  scattering_lengths=None):
+        """
+        Assign the scattering lengths either from a .lgt file, a user defined
```
andyfaff: It's not explained anywhere what an lgt file is.

refnx/reflect/_mdsimulation.py
```diff
@@ -0,0 +1,307 @@
...
+        array or allow it to be determined from the elements listed in the
```
andyfaff: And how is it determined from the elements listed in the PDB file?

refnx/reflect/_mdsimulation.py
andyfaff: (although it is below). Perhaps put a description in a `Notes` section.

refnx/reflect/_mdsimulation.py
```diff
@@ -0,0 +1,307 @@
...
+            Only required if the radiation is 'xray'. This is the energy of the
+            probing X-ray radiation, in units of kilo electron Volts
+        lgtfile: str, optional
+            The path and name of the .lgt file (if present), which contains the
+            scattering lengths of each of the atom type in the pdbfile.
+            Currently the .lgt file style that is supported is a 3 column space
+            separated txt file where the columns are atom_type,
+            real_scattering_length, and imaginary_scattering_length
+            respectively. If a lgtfile is not used the scattering lengths for
+            the system will be determined based on the element type that is
```
andyfaff: is that from periodictable?

refnx/reflect/_mdsimulation.py
andyfaff: Because it's not clear why the lgt file is required from the docstring, i.e. the coarse grained model.

refnx/reflect/_mdsimulation.py
```diff
@@ -0,0 +1,307 @@
...
+            defined in the final column of the pdb file.
+        atom_types: str, array_like
+            An array of length N, where N is the number of atom types in the
+            simulation trajectory, which defines the atom types to assign
+            scattering lengths.
+        scattering_lengths: float, array_like
```
andyfaff: Now I'm confused. Why is `scattering_lengths` required? Aren't these determined using `periodictable` or from the `lgt` file?

refnx/reflect/_mdsimulation.py
```diff
@@ -0,0 +1,307 @@
...
+            An array of shape [N, 2], where N is the number of atom types in
```
andyfaff: Presumably the first column is real, the second imaginary?

refnx/reflect/_mdsimulation.py
```diff
@@ -0,0 +1,307 @@
...
+        lgtfile: str, optional
+            The path and name of the .lgt file (if present), which contains the
+            scattering lengths of each of the atom type in the pdbfile.
+            Currently the .lgt file style that is supported is a 3 column space
+            separated txt file where the columns are atom_type,
+            real_scattering_length, and imaginary_scattering_length
+            respectively. If a lgtfile is not used the scattering lengths for
+            the system will be determined based on the element type that is
+            defined in the final column of the pdb file.
+        atom_types: str, array_like
```
andyfaff: what is `atom_types` used for?

refnx/reflect/_mdsimulation.py
andyfaff: is it an alternative to the lgt file? If all are specified which one wins out?

refnx/reflect/_mdsimulation.py
```diff
@@ -0,0 +1,307 @@
...
+                    self.scatlens[atom_types[i]] = scattering_lengths[i]
+            else:
+                raise ValueError('The lengths of the atom types must be the '
+                                 'same as the lengths of the pairs of '
+                                 'scattering lengths.')
+        elif lgtfile:
+            self.lgtfile = lgtfile
+            self.read_lgt()
+        else:
+            if radiation == 'neutron':
```
andyfaff: I think this should work.
```
self.neutron = radiation == 'neutron'
```

refnx/reflect/_mdsimulation.py
```diff
@@ -0,0 +1,307 @@
...
+    def read_pdb(self):
+        """Parse pdb file.
+
+        Parses the pdbfile.
+        """
+        parser = PDBParser()
+        self.structure = parser.get_structure('model', self.pdbfile)
+        file = open(self.pdbfile, 'r')
+        for line in file:
+            if line.startswith('CRYST1') and not np.any(self.dimensions):
```
andyfaff: Is this file parsing bit robust - is there any variation in PDF file that causes this to fall over?

refnx/reflect/_mdsimulation.py
arm61: No longer required, have detailed precedence

refnx/reflect/_mdsimulation.py
arm61: the atom_types are paired with the scattering lengths. Basically they are the names corresponding to the appropriate scattering length (maybe it would be better done with a `dict`?)

refnx/reflect/_mdsimulation.py
arm61: how would this would with the else statement below (I'm not familiar with that syntax)

refnx/reflect/_mdsimulation.py
arm61: The amount of variation in the pdb file format is sickening, to be honest (a headache for many people). That said, this is probably the most robust way possible, as often pdb files are output with things in incorrect columns, so the column approach that really should be used will fall down.

arm61:
Sorry for the delay in getting somethign more done. Was away at a conference

arm61:
Closed and reopened to get travis to run again to assess if the weird errors in build 902.2 were anomalous or not.

arm61:
@andyfaff any thoughts on the weird deprecation errors that are messing with the mac os travis build? They don't appear to be associated with anything that I have done, as far as I can tell

andyfaff:
Sorry, on holiday until mid Oct. I'll have a look as soon as I can. Latest
pytest seems to cause these issues.

arm61:
Sorry man!

Enjoy your holiday, a lot more important  :smile:

arm61:
Tried to determine if the error for due to python 3.7 (and it looks like it is). However the builds are now failing due to what looks like a circular dependency loop in the notebook due to the new quickplot stuff.

andyfaff:
@arm61 I've recently created a repository, [refnx-models](https://github.com/refnx/refnx-models), for user contributed code - similar to the sasview marketplace. I'm thinking that this code may be better off there than in the core package. The hurdles one has to jump to do that are probably lower, and it can get merged sooner.

arm61:
I have seen the refnx-models repo (plannig to put the volmono class in once the appropriate paper is published).

I don't mind some time being taken over getting the simulation stuff included in `refnx` if you think that it would be useful there (the paper that I have used it in is currently with collaborators so dunno when it will be submitted). If however you reckon that it is a bit out of scope of the refnx main code that is cool and I will move it over. 

Basically it is up to you where it fits in and I am not really in a rush (the only reason I am messing about with the pull request is to reduce the barrier when you have a chance).

arm61:
Gonna close this and instead put the code in the refnx-models. Once the paper is published in full that is.

PR #223, opened 2018-08-24
ENH: initial pnr calculation

andyfaff:
Still needs checking + more tests against established programs. I've verified that NSF works though.


andyfaff:
Note, this also bumps minimum python version to 3.5, as a consequence of the notation for matrix multiplication.

andyfaff:
@llimeht, any objections to dropping 2.7? I could restore it, but perhaps it makes sense to drop, given that refnx is a relatively new project.

PR #226, opened 2018-09-07
TST: check that diffev is repeatable

andyfaff:


andyfaff:
Darn, test fail because the PR for the DES context manager is not added yet.

PR #229, opened 2018-10-03
ENH: Mask data points

andyfaff:


andyfaff:
@tjmurdoch this will do masking. Just need to add a few more tests

PR #237, opened 2018-11-26
Tests GH236

andyfaff:
@igresh this test shows that the file can be loaded. I can put more code into the PR to fix the issue if I ever find out how to reproduce it.


andyfaff:
@igresh could you test this branch to see if the load works for you?

PR #245, opened 2019-01-14
MAINT: remove pandas as a mandatory dependency

andyfaff:
It's only needed for reduction, and it's noted in setup.py in the extras_require section.


andyfaff:
Merging. The fail on azure is caused by Theano and is noted at https://github.com/pymc-devs/pymc3/issues/3340. It should be fixed in theano master

PR #268, opened 2019-02-04
ENH: add __main__ to code fragment

andyfaff:


andyfaff:
This will enable users to export a code fragment, which is now fully executable, from the refnx gui. THe script will first do a differential evolution fit, then do MCMC sampling of the objective. One specifies the number of steps, walkers and thin values. The MCMC steps are saved to file, with the objective being printed to screen after the fit. It would be straightforward to amend the code fragment for an MPI environment.

PR #279, opened 2019-02-11
ENH: adds Spline to the qt gui

andyfaff:


andyfaff:
@igresh these are the changes you need to make to add a Component to the GUI.

PR #287, opened 2019-02-22
TST: compare c+py ref calc

andyfaff:
@mdoucet has created a repo comparing refl1d and refnx output for some corner cases. This PR checks the C + py abeles calc are the same in refnx for that corner case.

This PR also moves from the MyComplex class to the C++ implementation (less things to go wrong).


andyfaff:
Note, this PR also starts ignoring irho for fronting, which was being used. This is bring calculation into line with refl1d.abeles. It's possible, but v.v. rare that this would introduce a miscalculation into a user experiment.

PR #293, opened 2019-03-03
CI: azure artifacts

andyfaff:


andyfaff:
From now on the azure pipeline will make all the py3.x refnx wheels. Yipee, no longer do I have to make them.

PR #297, opened 2019-03-05
MAINT: speed up C implementation by choosing larger TINY number

andyfaff:


arm61:
Have you considered [numba](http://numba.pydata.org) as a further python speed up? I used it in [pylj](https://github.com/arm61/pylj/pull/40) pure python md steps became noticeably faster with it

andyfaff:
in #289 I transferred from using my own complex number library to using the C++ <complex.h> implementation. 

This was because the reflectivity calculation was using the wrong branch cut for a complex sqrt when there was non-zero absorption in middle layers, but the backing medium had an absorption of 0.
The effect would be negligible for most cases, resolving itself when the backing medium was given a tiny amount of absorption. The issue was fixed by giving all layers an extra amount of imaginary SLD [plus going to the <complex> library because my sqrt code wouldn't have made the same results for choosing the branch cut.  The output of sqrt(-4.0 + 0.0j) and sqrt(-4.0 - 0.0j) is the case in point].

It turns out the <complex> library is slower than my implementation, probably because it's more careful. The tiny amount I was adding on to the imaginary part was on the order of 2.e-308j. It turns out that it takes a much longer time to calculate a sqrt when the imaginary part is so small, 1e-30j is a lot faster and doesn't produce materially different output. Of course 1e-30j is next to zero when considering an absorption term.

This PR was making the change to 1e-30. It came about because I was profiling the reflectivity calculation in a Jupyter notebook.

 - The pure [Python implementation](https://github.com/refnx/refnx/blob/master/refnx/reflect/_reflect.py#L29) was taking around 433 us.
 - With a tiny value of 2e-308j the [C calculation](https://github.com/refnx/refnx/blob/master/src/_creflect.pyx#L25) was taking something like 700 us. Using a smaller tiny value resulted in a speed up to 254 us.
 - the alternate [cython implementation](https://github.com/refnx/refnx/blob/master/src/_creflect.pyx#L66) (unused) takes around 572 us.

For the reflectivity calculation the main bottlenecks are complex `sqrt` and complex `exp`, the matrix multiplication takes a bit of time too. Everything else pales into the background.
I tried using numba for doing the complex sqrt and complex exp in the pure Python implementation, I didn't get any speedup, so I didn't try using numba for the whole thing. The pure Python implementation is all vectorised with numpy, so one has to try really hard to improve on it. I've looked at all the implementations quite closely, but there's nothing that stands out as streamline-able. Can I encourage you to look at them and see if you spot anything?

Of course, one then has to be clever when doing the resolution smearing as well. [Point by point](https://github.com/refnx/refnx/blob/master/refnx/reflect/reflect_model.py#L487) resolution smearing is a whole lot slower than constant [dQ/Q](https://github.com/refnx/refnx/blob/master/refnx/reflect/reflect_model.py#L552). A big bottleneck then is the convolution and [interpolation](https://github.com/refnx/refnx/blob/master/refnx/reflect/reflect_model.py#L603). A second set of eyes would be welcome there as well. A great advantage of the code being in Python is that it's so easy to test things out. I would dread to be doing all this type of development in pure C/C++.

(Long reply because I'm writing it down for my own use later).

andyfaff:
I've put a gist of the alternative cython implementation [here](https://gist.github.com/andyfaff/56348832b9bf30b40a6cb1059adc95c6). Please feel free to mess around with it, and try numba on it, to see if speedups are possible.

PR #307, opened 2019-03-21
Cython

andyfaff:


andyfaff:
Pure cython with prange (_cyreflect.pyx) seems to be faster than farming out to the _creflect.pyx module (which then calls refcalc.cpp which uses pthreads). This PR is investigating how easy this is to do. However issues include:

- no openmp support on macOS. This has to be enabled by installing llvm-openmp and clang via conda, then setting the environment variables `CC=clang` and `CXX=clang`. One wouldn't to make a wheel on macOS with that because the openmp library may not be available.
- on macOS the openmp version crashes (if mkl numpy is installed) because there are two openmp libraries that are initialised. This can be circumvented by setting the `KMP_DUPLICATE_LIB_OK='True'` environment variable
- with a Docker Ubuntu Linux image the PR is currently failing. A multiprocessing plus parallelised calculation with `_creflect.abeles` works, but when a multiprocessing Pool calculation is done on `ReflectModel.model` the process hangs.

andyfaff:
It seems like the deadlocks I was experiencing on Linux are described at https://github.com/scikit-learn/scikit-learn/issues/7650

> At the moment it is not safe to use OpenMP for low overhead thread based parallelism in our Cython code base because of a bad interaction between multiprocessing.Pool (worker processes are necessarily created by fork without exec under Python 2) and the openmp runtime library libgomp used in GCC. This can cause the program to silently freeze.

Issue #318, opened 2019-05-09
CurveFitter.fit() only optimises nll


arm61:
As it stands the [`CurveFitter.fit()`](https://github.com/refnx/refnx/blob/ee4227275c0d2d2f414a09e5f478ee220cb9a687/refnx/analysis/curvefitter.py#L538) function will only minimise the negative log-likelihood (or residuals for least-squares). 

This means that for systems such as the [`LipidLeaflet`](https://github.com/refnx/refnx/blob/ee4227275c0d2d2f414a09e5f478ee220cb9a687/refnx/reflect/_lipid.py#L10) class, the [`logp()`](https://github.com/refnx/refnx/blob/ee4227275c0d2d2f414a09e5f478ee220cb9a687/refnx/reflect/_lipid.py#L270) function has no effect on the initial fitting. 

Would it be relevant to allow the `CurveFitter.fit()` function to optimise the negaive log posterior in addition to the negative log likelihood?

andyfaff:
If all the probability distributions for `logp` are uniform (majority of cases) then there is no difference to maximising the posterior compared to the likelihood, it's just a constant addition. They would start to differ when the priors aren't uniform. The `differential_evolution` and `LBFGS` algorithms are good for optimising parameter values within those bounds.
The main problem with minimising negative logposterior is that if there are some impossible regions (`logp == -np.inf`) then the optimizers may start to fall over. I think differential evolution would cope, but LBFGSB would fail.

arm61:
Interesting, I will test the method that I have implemented with a couple of different optimisation methods

andyfaff:
https://wiseodd.github.io/techblog/2017/01/01/mle-vs-map/

arm61:
I am wanting to move towards a Gaussian prior implementation (which I am aware would require some special case `bounds` being implemented). However, in the meantime it is mostly about getting the `-np.inf` to have an effect.

andyfaff:
You can use any scipy stats distribution with the `PDF` class, enabling a
Gaussian prior. This obviously works for the MCMC sampling, being encoded
in the log posterior. It should therefore also work for minimising negative
log posterior. I reckon it should be possible to use differential evolution
for that.

andyfaff:
The PDF class is [here](https://github.com/refnx/refnx/blob/master/refnx/analysis/bounds.py#L72).

arm61:
Resolved in #319

PR #319, opened 2019-05-21
A solution to minimise the nlpost

arm61:
A possible (pragmatic) solution to https://github.com/refnx/refnx/issues/318

### DIFF
--- a/refnx/analysis/curvefitter.py
+++ b/refnx/analysis/curvefitter.py
@@ -535,7 +535,7 @@ class CurveFitter(object):
         # sets parameter value and stderr
         return process_chain(self.objective, self.chain)
 
-    def fit(self, method='L-BFGS-B', **kws):
+    def fit(self, method='L-BFGS-B', prior=False, **kws):
         """
         Obtain the maximum log-likelihood estimate (mode) of the objective. For
         a least-squares objective this would correspond to lowest chi2.
@@ -555,6 +555,11 @@ class CurveFitter(object):
 
             You can also choose many of the minimizers from
             ``scipy.optimize.minimize``.
+        prior : bool
+            Have prior probabilities been defined that should be considered?
+            This is only supported for the differential_evolution optimisation
+            method, and if True will calculate the negative log-posterior
+            probability.
         kws : dict
             Additional arguments are passed to the underlying minimization
             method.
@@ -604,7 +609,16 @@ class CurveFitter(object):
         # bounds
         elif method in ['differential_evolution', 'dual_annealing', 'shgo']:
             mini = getattr(sciopt, method)
-            res = mini(self.objective.nll, **_min_kws)
+            if prior:
+                if method == 'differential_evolution':
+                    res = mini(self.objective.nlpost, **_min_kws)
+                else:
+                    raise ValueError('The use of a prior probability is '
+                                     'currently only supported for the '
+                                     'differential_evolution optimisation '
+                                     'method.')
+            else:
+                res = mini(self.objective.nll, **_min_kws)
         else:
             # otherwise stick it to minimizer. Default being L-BFGS-B
             _min_kws['method'] = method
--- a/refnx/analysis/curvefitter.py
+++ b/refnx/analysis/curvefitter.py
@@ -535,10 +535,11 @@ class CurveFitter(object):
         # sets parameter value and stderr
         return process_chain(self.objective, self.chain)
 
-    def fit(self, method='L-BFGS-B', prior=False, **kws):
+    def fit(self, method='L-BFGS-B', target='nll', **kws):
         """
-        Obtain the maximum log-likelihood estimate (mode) of the objective. For
-        a least-squares objective this would correspond to lowest chi2.
+        Obtain the maximum log-likelihood, or log-posterior, estimate (mode)
+        of the objective. For a least-squares objective maximum log-likelihood
+        corresponds to lowest chi2.
 
         Parameters
         ----------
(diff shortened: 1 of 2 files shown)


arm61:
Let me know thoughts on this style of solution and I will get some unit tests sorted.

refnx/analysis/curvefitter.py
```diff
@@ -535,7 +535,7 @@ def _callback_wrapper(state, h=None):
         # sets parameter value and stderr
         return process_chain(self.objective, self.chain)
 
-    def fit(self, method='L-BFGS-B', **kws):
+    def fit(self, method='L-BFGS-B', prior=False, **kws):
```
andyfaff: I'd prefer to use a *slightly* different approach. I'm not much good with naming, but how about we use a `target='nll'` keyword. The options for the `target` keyword would be `{'nll', 'nlpost'}`.
Default would be to do the maximum log-likelihood, but the other option would be maximise the log-posterior (often referred to as MAP).
I think that `nlpost` will fail if gradient based minimisers stray into an area for which the prior is non-finite (`-np.inf`), otherwise they should be fine. Differential evolution should be able to cope with non-finite priors so long as they're not ubiquitous.

I think the biggest challenge is the use of a `PDF` object to specify the prior/bounds for a given parameter. Let's say the `PDF` is Gaussian. It's easy to calculate the probabilities for that. The issue comes when an optimisation method (e.g. differential evolution) mandates lower and upper bounds for the parameter. The fitter [looks to each parameter](https://github.com/refnx/refnx/blob/master/refnx/analysis/curvefitter.py#L846) to see if its bounds attribute is an `Interval`. If it's not an `Interval`, which is won't be if you use a `PDF`, then the fitting process won't work because there will be no lower and upper bounds. I'm not sure how to get around this challenge, it'll require some thought.

TLDR: Not sure if it's possible to use anything other than a non-uniform distribution with a method that works on box (lower, upper) bounds.

refnx/analysis/curvefitter.py
```diff
@@ -555,6 +555,11 @@ def fit(self, method='L-BFGS-B', **kws):
 
             You can also choose many of the minimizers from
             ``scipy.optimize.minimize``.
+        prior : bool
```
andyfaff: `target : {'nll', 'nlpost'}, optional`

refnx/analysis/curvefitter.py
```diff
@@ -555,6 +555,11 @@ def fit(self, method='L-BFGS-B', **kws):
...
+            Have prior probabilities been defined that should be considered?
```
andyfaff: docstring amended accordingly.

refnx/analysis/curvefitter.py
andyfaff: It's only going to work on the scalar minimisers, not least-squares.

refnx/analysis/curvefitter.py
```diff
@@ -555,6 +555,11 @@ def fit(self, method='L-BFGS-B', **kws):
...
+            This is only supported for the differential_evolution optimisation
```
andyfaff: Should work for any of the scalar minimisers, but not least-squares.

refnx/analysis/curvefitter.py
```diff
@@ -604,7 +609,16 @@ def fit(self, method='L-BFGS-B', **kws):
         # bounds
         elif method in ['differential_evolution', 'dual_annealing', 'shgo']:
             mini = getattr(sciopt, method)
-            res = mini(self.objective.nll, **_min_kws)
+            if prior:
```
andyfaff: These additions aren't needed.

refnx/analysis/curvefitter.py
```diff
@@ -604,7 +609,16 @@ def fit(self, method='L-BFGS-B', **kws):
...
+                if method == 'differential_evolution':
+                    res = mini(self.objective.nlpost, **_min_kws)
+                else:
+                    raise ValueError('The use of a prior probability is '
+                                     'currently only supported for the '
+                                     'differential_evolution optimisation '
+                                     'method.')
+            else:
+                res = mini(self.objective.nll, **_min_kws)
```
andyfaff: All you need to do is:

```
cost = self.objective.nll
if target == 'nlpost':
    cost = self.objective.nlpost

res = mini(cost, **_min_kws)
```

andyfaff:
There will be challenges for non-uniform priors for the scalar minimisers which use box (lower, upper) bounds. However, I'd make the changes requested here and we can cross that bridge when we come to it.

andyfaff:
I have an idea of how to approximate bounds for a MAP. `scipy.stats` distributions have various methods, one of which is the [`ppf` function](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.rv_continuous.ppf.html).

We could estimate that the lower and upper bound is at e.g. 0.01 and 0.99 of the cumulative distribution function (CDF). For example, the following code works out the x-values for 16 % and 84% of the way through the normal distribution. These should be -1 and 1 respectively because those values lie 1 s.d. away from the mean.

```
>>> from scipy.stats import norm
>>> print(norm.ppf([0.16, 0.84]))
[-0.99445788  0.99445788]
```

Thus, the code in the `fit` method could be:

- see if it's an nlpost or nll fit.
- Go through each of the parameters and ensure they all have `Bound`s. If the `Bound`s are `Interval` it's easy to setup lower and upper limit. If the `Bound` is a `PDF` then estimate a lower and upper limit by trying to use `PDF.ppf` with e.g. 0.01 and 0.99 of the CDF.

This will fail to find the maximum probability if the value lies in the tails though.

arm61:
This is a nice general solution!

Sorry for the delay in sorting it myself (I had my viva on Thursday and have been ignoring work since then) but I think that your solution might be more general than the method I had pieced together.

PR #332, opened 2019-06-06
ENH: specify Mapwrapper context

andyfaff:


andyfaff:
Note that multiprocessing on macOS, mixed with qt GUI work, means that it's a better idea to use 'spawn' than 'fork'

PR #364, opened 2019-08-22
MAINT: retro add logprob attribute when loading mtft file

andyfaff:


andyfaff:
Added a `_logprob` attribute to `bounds.Interval`. When loading an experiment file from a previous version of refnx this attribute isn't present in the pickle. Add it back in...

PR #369, opened 2019-09-21
Universal event file reader

andyfaff:
The event file reader will now read any ANSTO packedbin format.


andyfaff:
@geishm-ansto, thank you so much for pointing to that header file. I used it to make this PR, which is a universal Python reader for event files. The one thing the original header file didn't have was a mechanism for converting uint32 --> int32 for detectors whose x and y pixels are signed (`evt_stg_xy_signed == 1`).

geishm-ansto:
I’m happy it worked out.

Regards

PR #371, opened 2019-09-29
Bound from rvcontinuous

arm61:
If `a` and `b` are defined for a `scipy.stats.rv_continuous` object use these as bounds

https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.rv_continuous.html

### DIFF
--- a/refnx/analysis/curvefitter.py
+++ b/refnx/analysis/curvefitter.py
@@ -9,6 +9,7 @@ import numpy as np
 from scipy._lib._util import check_random_state
 from scipy.optimize import minimize, differential_evolution, least_squares
 import scipy.optimize as sciopt
+import scipy.stats as scistat
 
 from refnx.analysis import Objective, Interval, PDF, is_parameter
 from refnx._lib import (unique as f_unique, MapWrapper,
@@ -907,6 +908,12 @@ def bounds_list(parameters):
                 isinstance(param.bounds, Interval)):
             bnd = param.bounds
             bounds.append((bnd.lb, bnd.ub))
+        elif (hasattr(param, 'bounds') and isinstance(param.bounds, PDF) and
+              isinstance(param.bounds.rv,
+                         (scistat.rv_continuous,
+                          scistat._distn_infrastructure.rv_frozen)) and
+              np.isfinite([param.bounds.rv.a, param.bounds.rv.b]).all()):
+            bounds.append((param.bounds.rv.a, param.bounds.rv.b))
         elif (hasattr(param, 'bounds') and isinstance(param.bounds, PDF) and
               hasattr(param.bounds.rv, 'ppf')):
             bounds.append(param.bounds.rv.ppf([0.005, 0.995]))
--- a/refnx/analysis/curvefitter.py
+++ b/refnx/analysis/curvefitter.py
@@ -909,11 +909,8 @@ def bounds_list(parameters):
             bnd = param.bounds
             bounds.append((bnd.lb, bnd.ub))
         elif (hasattr(param, 'bounds') and isinstance(param.bounds, PDF) and
-              isinstance(param.bounds.rv,
-                         (scistat.rv_continuous,
-                          scistat._distn_infrastructure.rv_frozen)) and
-              np.isfinite([param.bounds.rv.a, param.bounds.rv.b]).all()):
-            bounds.append((param.bounds.rv.a, param.bounds.rv.b))
+              np.isfinite([param.bounds.lb, param.bounds.ub]).all()):
+            bounds.append((param.bounds.lb, param.bounds.ub))
         elif (hasattr(param, 'bounds') and isinstance(param.bounds, PDF) and
               hasattr(param.bounds.rv, 'ppf')):
             bounds.append(param.bounds.rv.ppf([0.005, 0.995]))
--- a/refnx/analysis/bounds.py
+++ b/refnx/analysis/bounds.py
@@ -94,11 +94,13 @@ class PDF(Bounds):
     (-11.043938533204672, -11.043938533204672)
 
     """
-    def __init__(self, rv, seed=None):
+    def __init__(self, rv, bounds=[-np.inf, np.inf], seed=None):
         super(PDF, self).__init__(seed=seed)
         # we'll accept any object so long as it has logpdf and rvs methods
         if hasattr(rv, 'logpdf') and hasattr(rv, 'rvs'):
             self.rv = rv
+            self.lb = bounds[0]
+            self.ub = bounds[1]
         else:
             raise ValueError("You must initialise PDF with an object that has"
                              " logpdf and rvs methods")
--- a/refnx/analysis/test/test_curvefitter.py
+++ b/refnx/analysis/test/test_curvefitter.py
@@ -79,6 +79,9 @@ class TestCurveFitter(object):
         self.p[0].bounds = PDF(norm(0, 1))
         assert_allclose(bounds_list(self.p),
                         [norm(0, 1).ppf([0.005, 0.995]), (-100, 100)])
+        self.p[0].bounds.rv.a = 0.2
+        self.p[0].bounds.rv.b = 0.8
+        assert_allclose(bounds_list(self.p), [[0.2, 0.8], (-100, 100)])
 
     def test_constraints(self):
         # constraints should work during fitting
--- a/refnx/analysis/test/test_curvefitter.py
+++ b/refnx/analysis/test/test_curvefitter.py
@@ -3,7 +3,7 @@ import pickle
 
 import numpy as np
 import scipy.optimize as sciopt
-from scipy.stats import norm
+from scipy.stats import norm, truncnorm
 
 import pytest
 from numpy.testing import (assert_, assert_almost_equal, assert_equal,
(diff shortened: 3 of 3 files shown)


andyfaff:
This isn't quite right. `a` and `b` are shape parameters, and are used for many different continuous distributions. For example [genexpon](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.genexpon.html#scipy.stats.genexpon) has `a`, `b`, and `c` shape parameters and it's nowhere near a uniform distribution. So using `a` and `b` is not a good way of checking if `rv` is a uniform distribution.
I'm not sure if it's all that easy to check if `rv` is a uniform distribution either (can be frozen, or unfrozen, or a different kind of object implementing the logpdf and rvs methods). This is why the next line checks for the `ppf` method. This works for any `rv_continuous` and would set an interval corresponding to a CDF range of [0.005, 0.995].

arm61:
Not trying to check if rv is a uniform distribution. The `rv_continuous` gives `a` and `b` as the lower and upper bound of support for the distribution. 

The reason for this is I was using a truncated normal distribution as the prior but the `ppf` method meant that the bounds were substaintially closer than I wants (I wanted the `logp` to influence the fitting without the hards bounds from the `ppf` present, instead using the limits of truncation from the prior as the bounds). 

I hope that makes sense, I imagine that my method isn't perfect but it is working for my specific problem. It might be easier if there is a way to identify the particular distribution in use? And if it is a truncated function then used the bounds of truncation.

andyfaff:
Sorry. I learnt something today. I was getting confused because `a`, `b`, `c` are used as shape parameter names. However, they are also attributes of the `rv_continuous` object. The scipy documentation could do with improvement in that regard.

arm61:
No worries. Just wanna make sure that this won't cause any problems, but it seems that, for example, `a`, `b` and `c` are input args for the `genexpon` but they value of say `genexpon(1, 2, 3).a` is 0.

arm61:
Interestingly for `truncnorm` the values for `a` and `b` default to `-np.inf` regardless of the input values. Gonna see if I can make a pull request to Scipy to change that as it seems counter-intuitive

refnx/analysis/curvefitter.py
```diff
@@ -907,6 +907,9 @@ def bounds_list(parameters):
                 isinstance(param.bounds, Interval)):
             bnd = param.bounds
             bounds.append((bnd.lb, bnd.ub))
+        elif (hasattr(param, 'bounds') and isinstance(param.bounds, PDF) and
+              param.bounds.rv.a != -np.inf and param.bounds.rv.b != -np.inf):
```
andyfaff: PDF can be created with any class that has `logpdf` and `rvs` methods. So perhaps this code block is better written as:

```
        elif (hasattr(param, 'bounds') and isinstance(param.bounds, PDF) and isinstance(param.bounds.rv, (scipy.stats.rv_continuous, scipy.stats._distn_infrastructure.rv_frozen)) and np.isfinite([param.bounds.rv.a, param.bounds.rv.b]).all()):
```

assuming it works.

andyfaff:
I'd expect truncnorm.a to be the left truncation value, etc.

arm61:
> I'd expect truncnorm.a to be the left truncation value, etc.

Same, however for `truncnorm` these default to `-np.inf`, therefore I am currently assigning `a` and `b` after the fact. That is a Scipy issue though so I will have a look over there today

andyfaff:
```
>>> import numpy as np
>>> from scipy.stats import truncnorm
>>> rv = truncnorm(-5, 5)
>>> rv.a
-5
>>> rv.a = -6
>>> rv.pdf(-5.5)
0.0
```

So `rv.a` is fine to start with. However, changing the rv.a after a distribution has been constructed doesn't work. In that case `rv.a` should be a property that isn't settable.

andyfaff:
This means the test in this PR isn't the best. THe test should be create a `PDF` with `truncnorm`, then to figure out if those bounds are set correctly.

arm61:
Although this introduces another problem in that `a` and `b` in the `truncnorm` are set relative to the mean and therefore don't match up with the bounds when the `mean != 0`.

andyfaff:
the PR needs to be changed to account for that.

arm61:
> the PR needs to be changed to account for that.

I'll get back to you with something that works better later today

arm61:
Okay it seems like this is a harder problem than expected, as it is not possible to recover the mean and std of the parent normal distribution from a truncated normal distribution without having the parent. distribution apparently...

I have committed something that is a bit of a change, when a PDF is used, there is now an option of `bounds` which sets hards bounds in the `bounds_list`. 

I know that this is quite different to what I started out with but I can't come up with a cleaner way to get what I am needing.

arm61:
closed and reopened to get the CI to run again because it seemed that the azure build died for Tcl reasons and I couldn't see how to restart the build.

andyfaff:
Back to work after holiday now. It's still not clear to me that this PR is necessary.

- `bounds_list` is only used if you are using `CurveFitter.fit`, it's immaterial to MCMC sampling
- `bounds_list` is used to create lower and upper box bounds for those fitting methods that otherwise wouldn't work. If
- it covers 99% of the statistical distribution, so I would've thought that the True solution would very likely lie in between those bounds (esp. for a normal distribution).
- the bounds created by `bounds_list` have no effect on either the `nll` or `nlpost` statistics that are used in fit. `nll` is the negative log-likelihood, so there's no impact of the prior distribution there. `nlpost` is the negative log-posterior probability. The log-prior does contribute to this statistic, but it's contribution is solely determined by `Objective.logp`. In other words the True log-prior for `truncnorm` is used, and is unaffected by the bounds created by `bounds_list`.

The only reason to expand `bounds_list` from it's current implementation is if the True solution is likely to lie outside the 99% range. We could always expand that to 99.9% if that's likely to be the case.

arm61:
this might make my case redundant

> bounds_list is only used if you are using CurveFitter.fit, it's immaterial to MCMC sampling

However, how then is a bound imposed in the MCMC sampling (is this translated to a uniform prior?)

andyfaff:
If a `Parameter` is varying, and it has no specified bounds, then the default lower and upper bound is `[-np.inf, np.inf]`. It's contribution to the log-prior term is always (uniformly) zero. This is not an issue for the MCMC sampling, the walkers are allowed to access any region which has higher log-posterior probabilities. Obviously one can't initialise MCMC walkers by sampling the prior for such a parameter. The only ways to do this is via jittering or estimating from the local covariance matrix.

Fitting is a different story. For example the differential_evolution implementation requires finite lower and upper bounds to work, an Exception is raised otherwise. However, if the user has gone to the trouble of providing a PDF for the bounds, then bounds_list tries to estimate what a sensible lower and upper bound could be using `rv_continuous.ppf`. Whether or not that affects the fit or not depends on whether one is using the `nll` or `nlpost` option for `Curvefitter.fit`.

arm61:
Maybe I am not getting something, but surely if 

> the walkers are allowed to access any region which has higher log-posterior probabilities

Then we should *never* see the thing in a `corner` plot like that in Figure S19 (Page 13) of this http://www.rsc.org/suppdata/c9/cp/c9cp00203k/c9cp00203k1.pdf. In particular the roughness parameter, where the lower bound of the roughness was a value of 3.3 Å (imposed at the insistance of others). 

Of is it the case the outside the bounds the `logl` is `-np.inf` ?

andyfaff:
Definitely something missing from your understanding. Consider the following cases, all of which are uniform priors (an Interval bound):

- [-np.inf, np.inf], log-prior=0
- [-np.inf, finite], log-prior=0
- [finite, np.inf], log-prior=0
- [lb, ub], log-prior=log(1/(ub-lb))

For all the cases if you're outside those finite limits then the log-prior is impossible, -np.inf.

The walkers are free to explore any space which has `probability > -np.inf`.

arm61:
Alright that makes sense, however this then means that the bounds **do** influence the MCMC. 

So by applying bounds (which in MCMC are still hard bounds as outside them `-np.inf` is enforced) of 99 % of the statistical distribution you are implying that that 1 % is *impossible* (`-np.inf`). What this PR is doing, is letting the user define what is possible (albeit improbable). 

Hope that makes sense, the motivation for this PR in the first place was that I was finding the bounds imposed on the statistical distribution (99 %) to be too constrictive and wanting the MCMC to be able to investigate the *improbable* region, if it saw fit.

andyfaff:
> that the bounds do influence the MCMC.

Yes, all the `Parameter.bounds` distributions are used 'as is' in the MCMC sampling. There is no 99% restriction in the MCMC sampling, it samples 100% of the probable region.

> So by applying bounds (which in MCMC are still hard bounds as outside them -np.inf is enforced) of 99 % of the statistical distribution you are implying that that 1 % is impossible (-np.inf).

The *`bounds_list` function that figures out the 99% of the distribution is not used for MCMC at all*. It's only usage is in the [`fit` method](https://github.com/refnx/refnx/blob/master/refnx/analysis/curvefitter.py#L586). differential_evolution needs finite lower and upper bounds to work. Say for example one parameter has a `scipy.stats.norm` for a bound. To enable fitting with DE for that parameter we need to estimate a finite lower and upper bound. This is done by estimating 99% of the distribution.

- MCMC sampling (`Curvefitter.sample`) uses 100% of the probable region specified by Parameter.bound.
- Fitting (`Curvefitter.fit`) uses 100% of the probable region specified by bounds that are described by a uniform distribution, and 99% of the probable region specified by bounds that are described by a PDF (e.g. `scipy.stats.norm`).

arm61:
Alright, closing for now as you are right that it is unnecessary!

andyfaff:
Thanks for the discussion though. It's been a learning experience on both sides.

arm61:
definitely! :metal:

PR #373, opened 2019-10-26
DOC: expand dependency list

andyfaff:
A user reports that they couldn't start the refnx gui with the suggested way of setting up the conda environment (in the installation instructions documentation).

This PR adds a few packages to the suggested setup that will allow the refnx gui to start from the command line. These packages are not required if the GUI isn't being used, but it probably makes sense to have them there. @joscooper

### DIFF
--- a/doc/installation.rst
+++ b/doc/installation.rst
@@ -35,7 +35,7 @@ Creating a conda environment
 
     ::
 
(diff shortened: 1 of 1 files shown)


andyfaff:
Failing test entry on travis is unrelated to change, something has changed on the CI.

PR #403, opened 2020-01-18
BUG: dragging components in a stack

andyfaff:


andyfaff:
Failure is due to Ubuntu not finding apt packages

PR #463, opened 2020-06-04
TST: reduction makes use of refnx-testdata

andyfaff:


andyfaff:
This should shrink the size of the sdist considerably

PR #464, opened 2020-06-05
MAINT: compensate for Parameter._stderr

andyfaff:


andyfaff:
Also adjust view.py with black.

PR #484, opened 2020-07-17
MAINT: patch _random_state

andyfaff:


andyfaff:
Darn - there are still scipy.stats distributions that aren't compatible with `Generator`.

PR #492, opened 2020-08-07
MAINT: fix DeprecationWarnings

andyfaff:
Removes a deprecation warning in mpl 3.3


andyfaff:
And also in pymc3 use.

PR #496, opened 2020-09-23
Pnr catalogue

ohcpaull:
Polarised catalogue that reads PNR-specific metadata which can help users find info about sample environments and flippers for specific scans. Have made it such that the class is implemented after the PlatypusCatalogue has been called - this might be able to be improved.

Feedback on the relevance of my commits for adding c files etc is appreciated.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -291,3 +291,4 @@ in 0.1.13.
 - Update emcee vendored code.
 - Fix repr of Model
 - Fix several warnings during test run.
+- Add error bar plotting to refnx GUI.
--- a/refnx/reflect/_app/graphproperties.py
+++ b/refnx/reflect/_app/graphproperties.py
@@ -76,7 +76,10 @@ class GraphProperties(dict):
         # pass
         if self.ax_data is not None:
             for key in _requiredgraphproperties:
-                self["data_properties"][key] = artist.getp(self.ax_data, key)
+                # ax_data is an ErrorbarContainer
+                self["data_properties"][key] = artist.getp(
+                    self.ax_data[0], key
+                )
 
         if self.ax_fit is not None:
             for key in _requiredgraphproperties:
--- a/refnx/reflect/_app/test/test_graphproperties.py
+++ b/refnx/reflect/_app/test/test_graphproperties.py
@@ -8,8 +8,8 @@ def test_graphproperties():
     gp = GraphProperties()
     fig = Figure()
     axes = fig.add_axes([0.06, 0.15, 0.9, 0.8])
-    line = axes.plot([0, 1, 2.0], [0, 1, 2.0])
-    gp.ax_data = line[0]
+    ebc = axes.errorbar([0, 1, 2.0], [0, 1, 2.0], [0, 1, 2.0])
+    gp.ax_data = ebc
     gp.save_graph_properties()
 
     pkl = pickle.dumps(gp)
--- a/refnx/reflect/_app/view.py
+++ b/refnx/reflect/_app/view.py
@@ -213,7 +213,8 @@ class MotofitMainWindow(QtWidgets.QMainWindow):
         # then try and load urls as data files
         try:
             loaded_names = self.load_data(urls_as_files)
-        except Exception:
+        except Exception as e:
+            print(e)
             loaded_names = []
 
         # then try and load the remainder as models
@@ -2239,24 +2240,27 @@ class MyReflectivityGraphs(FigureCanvas):
                     yt, edata = transform(dataset.x, dataset.y, dataset.y_err)
 
                 # add the dataset
-                line_instance = self.axes[0].plot(
+                error_bar_container = self.axes[0].errorbar(
                     dataset.x,
                     yt,
+                    yerr=edata,
                     markersize=3,
                     marker="o",
                     linestyle="",
                     markeredgecolor=None,
                     label=dataset.name,
                 )
-                line_instance[0].set_pickradius(5)
-                mfc = artist.getp(line_instance[0], "markerfacecolor")
-                artist.setp(line_instance[0], **{"markeredgecolor": mfc})
+                line_instance = error_bar_container[0]
+                line_instance.set_pickradius(5)
+                mfc = artist.getp(line_instance, "markerfacecolor")
+                artist.setp(line_instance, **{"markeredgecolor": mfc})
 
-                graph_properties["ax_data"] = line_instance[0]
-                if graph_properties["data_properties"]:
+                graph_properties["ax_data"] = error_bar_container
+                data_properties = graph_properties["data_properties"].copy()
+                if data_properties:
+                    artist.setp(error_bar_container[0], **data_properties)
                     artist.setp(
-                        graph_properties.ax_data,
-                        **graph_properties["data_properties"]
+                        error_bar_container[-1], color=data_properties["color"]
                     )
 
             yfit_t = data_object.generative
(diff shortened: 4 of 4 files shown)


refnx/reduce/platypusnexus.py
```diff
@@ -498,6 +498,142 @@ def _chopper_values(self, h5data):
...
+
+        d = self.cat
+
+        d = self._polarised_flippers(d, h5d)
+        d = self._analyser_flippers(d, h5d)
+        d = self._analyser_params(d, h5d)
+        d = self._check_sample_environments(d, h5d)
+
+    def _polarised_flippers(self, d, h5d):
+        d["pol_flip_freq"] = h5d[
```
andyfaff, line 533: how much of `_polarised_flippers` is required? I can see how it would be important to know if the flippers are on/off for reduction purposes, but does one need to know the current/frequency/voltage/guide/voltage?

refnx/reduce/platypusnexus.py
```diff
@@ -498,6 +498,142 @@ def _chopper_values(self, h5data):
...
+        for reflectometry reduction
+        Access information via dict access, e.g. cat['pol_flip_freq'].
+
+        Parameters
+        ----------
+        h5d - HDF5 file handle
+        """
+        super(PolarisedCatalogue, self).__init__(h5d)
+        # Is there a magnet?
+        self.ismagnet = False
```
andyfaff, line 519: ismagnet --> is_magnet

refnx/reduce/platypusnexus.py
```diff
@@ -498,6 +498,142 @@ def _chopper_values(self, h5data):
...
+        # Is there a cryocooler?
+        self.iscryo = False
```
andyfaff, line 521: is_cryo

refnx/reduce/platypusnexus.py
```diff
@@ -498,6 +498,142 @@ def _chopper_values(self, h5data):
...
+        # Is there a power supply?
+        self.isPS = False
```
andyfaff, line 523: what's the power supply for, is that a power supply for sample environment?

refnx/reduce/platypusnexus.py
```diff
@@ -498,6 +498,142 @@ def _chopper_values(self, h5data):
...
+        d["pol_flip_status"] = h5d[
+            "entry1/instrument/polarizer_flipper/flip_on"
+        ][0]
+        d["pol_guide_current"] = h5d[
+            "entry1/instrument/polarizer_flipper/guide_current"
+        ][0]
+        return d
+
+    def _analyser_flippers(self, d, h5d):
+        d["anal_flip_freq"] = h5d[
```
andyfaff, line 551: see comment for polarizer_flipper.

refnx/reduce/platypusnexus.py
```diff
@@ -498,6 +498,142 @@ def _chopper_values(self, h5data):
...
+        # Is there a magnet?
+        self.ismagnet = False
+        # Is there a cryocooler?
+        self.iscryo = False
+        # Is there a power supply?
+        self.isPS = False
+
+        d = self.cat
+
+        d = self._polarised_flippers(d, h5d)
```
andyfaff, line 527: --> `_polariser_flippers`

refnx/reduce/platypusnexus.py
```diff
@@ -498,6 +498,142 @@ def _chopper_values(self, h5data):
...
+        d = self._analyser_flippers(d, h5d)
+        d = self._analyser_params(d, h5d)
+        d = self._check_sample_environments(d, h5d)
+
+    def _polarised_flippers(self, d, h5d):
```
andyfaff, line 532: `_polariser_flippers`

refnx/reduce/platypusnexus.py
```diff
@@ -498,6 +498,142 @@ def _chopper_values(self, h5data):
...
+        d["anal_flip_status"] = h5d[
+            "entry1/instrument/analyzer_flipper/flip_on"
+        ][0]
+        d["anal_guide_current"] = h5d[
+            "entry1/instrument/analyzer_flipper/guide_current"
+        ][0]
+        return d
+
+    def _analyser_params(self, d, h5d):
+        d["analyzer_base"] = h5d["entry1/instrument/polarizer/anal_base"][0]
```
andyfaff, line 569: Are these really needed to be loaded for data reduction purposes? They're important to know for acquisition, but I can't see the value otherwise.

refnx/reduce/platypusnexus.py
```diff
@@ -498,6 +498,142 @@ def _chopper_values(self, h5data):
...
+            self.iscryo = True
+        except KeyError:
+            # Temperature sensor not used in measurement - set to None
+            d["temp_sensorA"] = None
+            d["temp_sensorB"] = None
+            d["temp_sensorC"] = None
+            d["temp_sensorD"] = None
+            d["temp_setpt1"] = None
+            d["temp_setpt2"] = None
+            self.iscryo = False
```
andyfaff, line 595: is_cryo

refnx/reduce/platypusnexus.py
```diff
@@ -498,6 +498,142 @@ def _chopper_values(self, h5data):
...
+
+        try:
+            # Try adding voltage supply to dict
+            d["pow_supply_volts"] = h5d["entry1/sample/power_supply/voltage"][
+                0
+            ]
+            d["pow_supply_current"] = h5d["entry1/sample/power_supply/amps"][0]
+            d["pow_supply_relay"] = h5d["entry1/sample/power_supply/relay"][0]
+            self.isPS = True
```
andyfaff, line 604: is_PS, but it's not clear what PS is. Perhaps is_power_supply?

refnx/reduce/platypusnexus.py
```diff
@@ -498,6 +498,142 @@ def _chopper_values(self, h5data):
...
+                "entry1/sample/ma1/sensor/nominal_outp_current"
+            ][0]
+            self.ismagnet = True
+        except KeyError:
+            # Magnetic field sensor not used in measurement - set to None
+            d["magnet_current_set"] = None
+            d["magnet_set_field"] = None
+            d["magnet_measured_field"] = None
+            d["magnet_output_current"] = None
+            self.ismagnet = False
```
andyfaff, line 633: is_magnet

andyfaff (changes requested):
I cleaned up the git commits and removed somethings that shouldn't have been added (e.g. changes to .pyd, .gitignore).

refnx/reduce/platypusnexus.py
ohcpaull: For normal usage, these parameters aren't very necessary - would mainly be for debugging if a scan went wrong for an unknown reason or for optimising the flippers by changing the frequency (which is done very rarely). This could be done otherwise in nexus file-reading software and we can trim these off if you prefer to keep the dict to the most important params

refnx/reduce/platypusnexus.py
ohcpaull: Yes, a power supply to apply voltages inside the cryostat

refnx/reduce/platypusnexus.py
ohcpaull: No, I don't think any of these are needed very often at all - this whole function can be deleted.

refnx/reduce/platypusnexus.py
ohcpaull: Yes this is correct

PR #502, opened 2020-10-20
Update bounds.py

arm61:
For the `invcdf` method of the `PDF` class, the `rv` object needs a `ppf` method. This previously was not reflected in the `__init__`.

### DIFF
--- a/refnx/analysis/bounds.py
+++ b/refnx/analysis/bounds.py
@@ -117,12 +117,12 @@ class PDF(Bounds):
     def __init__(self, rv):
         super(PDF, self).__init__()
         # we'll accept any object so long as it has logpdf and rvs methods
-        if hasattr(rv, "logpdf") and hasattr(rv, "rvs"):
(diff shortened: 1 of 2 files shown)


andyfaff:
The `invcdf` method was added so that it was easier to use dynesty; it's not really used elsewhere, unlike `rv.logpdf` and `rv.rvs`.

To use `invcdf` the `PDF.rv` object has to have a `ppf` method. This will be the case for all `scipy.stats.rv_continuous` distributions. If one isn't using dynesty (and therefore probably not using `PDF.invcdf`) is it important to check that the ppf method is implemented? If someone tries to use that method then all they'll see is an exception. I probably don't mind either way.

andyfaff:
There seems something going on with GH actions, hopefully it'll resolve itself soon.

arm61:
Cheers for the pull. 

Aware that this "fix" is a bit niche, but it was a problem that we (a student and myself) just fell into 'cause we are using a custom prior class and doing evidence estimation. So I thought a fix to acknowledge it might be useful 😄

PR #516, opened 2020-11-16
ENH: use vfsolv when makeing Slab from Scatterer

andyfaff:


andyfaff:
Please forgive for the typo in the commit message.

Issue #522, opened 2020-12-18
Suggestion: add the option to ignore zeros in ReflectDataset?


James-Durant:
Hi there,

When trying to fit data with points of zero reflectivity, understandably, NaNs are produced in the `Objective.logl` method and the following runtime error is raised:
![image](https://user-images.githubusercontent.com/43757679/102608868-79050e00-4122-11eb-9a4c-c4ce02bbf4d7.png)

Perhaps adding a method to the `ReflectDataset` class to ignore zeros or to remove them would be beneficial?

I'm currently using the following as a workaround:
```
dataset = np.loadtxt("test.dat")
dataset = dataset[(dataset != 0).all(1)]  #Remove any points with 0 reflectivity
q, r, r_error = dataset[:,0], dataset[:,1], dataset[:,2]
data = ReflectDataset([q, r, r_error])
```

Thanks,
James

arm61:
Check out the `mask` property in the `Data1D` class. It might allow for a shorter workaround?

James-Durant:
Ah yes, I forgot about that! Something like this works nicely:
```
data = ReflectDataset("test.dat") 
data.mask = (data.y != 0)
```

I raised the issue as, for someone new to refnx, it might not necessarily be obvious why they could be getting NaNs from fitting their data.

Anyway thank you for the suggestion Andrew,
James

arm61:
Potentially worth making a PR where an error is raised if any of the values or uncertainties in a dataset are 0, with the recommendation to apply a `mask`?

andyfaff:
James-Durant, thank you for contributing to refnx for the first time.

Experimental reflectivity curves may have points that have R=0, or even R < 0. With a perfect background subtraction one would expect half the points in that region to be above 0, and half below. Note - I would expect `sigma_R`, the uncertainty on the data points to be strictly greater than 0, `sigma_R > 0`.

If no `Transform` is used (fitting linY vs X), then this should be fine. However, there will be issues experienced if a logY vs X transform is used, as it can't cope with R <=0. 

People frequently use linY vs X, and it's therefore the wrong thing to do to filter those datapoints when they're loaded. It happens so frequently that it might be annoying to raise a warning when the dataset is loaded.

[...]

PR #524, opened 2020-12-21
Added runtime error to Transform in objective.py

James-Durant:
Added a `RuntimeError` to `Transform` for non-finite transformed values when using the logY transform. This should address issue #522.


andyfaff:
I would use `if not np.isfinite(arr).all()` after the transform.

James-Durant:
Ah good suggestion! I'll update my PR to use that

andyfaff (approved):
Looks good. Thank you for your contribution James.

PR #533, opened 2021-02-02
Spinset classes

ohcpaull:
Adding classes for SpinChannel and SpinSet. 

SpinChannel is an Enum class specifying the spin state in a PlatypusNexus file/

SpinSet is a set of up to 4 spin channels that are kept together for polarisation efficiency correction as well as for use in a future PolarisedReduce class for reducing multiple spin-polarised datasets.

### DIFF
--- a/refnx/reduce/__init__.py
+++ b/refnx/reduce/__init__.py
@@ -14,6 +14,8 @@ from refnx.reduce.platypusnexus import (
     Catalogue,
     SpatzNexus,
     ReductionOptions,
+    SpinChannel,
+    SpinSet,
 )
 from refnx.reduce.batchreduction import BatchReducer
 from refnx.reduce.xray import reduce_xrdml
--- a/refnx/reduce/platypusnexus.py
+++ b/refnx/reduce/platypusnexus.py
@@ -9,6 +9,7 @@ from time import gmtime, strftime
 import string
 import warnings
 from contextlib import contextmanager
+from enum import Enum
 
 from scipy.optimize import leastsq, curve_fit
 from scipy.stats import t
@@ -624,6 +625,166 @@ class PolarisedCatalogue(PlatypusCatalogue):
             d["magnet_output_current"] = None
             self.is_magnet = False
         return d
+
+
+class SpinChannel(Enum):
+    """
+    Describes the spin state of a polarised neutron beam.
+    """
+
+    UP_UP = (1, 1)
+    UP_DOWN = (1, 0)
+    DOWN_UP = (0, 1)
+    DOWN_DOWN = (0, 0)
+
+
+class SpinSet(object):
+    """
+    Describes a set of spin-channels at a given angle.
+
+    TODO: implement polarisation efficiency correction within this class
+
+    Parameters
+    ----------
+    list of {str, h5data}
+        list of str, or list of h5py file handles pointing to
+        a set of polarised neutron beam files
+
+    data_folder: {str, Path}
+        Path to the data folder containing the data to be reduced.
+    """
+
+    def __init__(
+        self, down_down, up_up, down_up=None, up_down=None, data_folder=None
+    ):
+        # Currently only Platypus has polarisation elements
+        self.reflect_klass = PlatypusNexus
+        self.data_folder = data_folder
+
+        channels = [down_down, up_up, down_up, up_down]
+
+        # initialise spin channels
+        self.dd = self.du = self.ud = self.uu = None
+
+        # initialise reduction options for each spin channel
+        reduction_options = {
+            "lo_wavelength": 2.5,
+            "hi_wavelength": 12.5,
+            "rebin_percent": 3,
+        }
+        self.dd_opts = reduction_options.copy()
+        self.du_opts = reduction_options.copy()
+        self.ud_opts = reduction_options.copy()
+        self.uu_opts = reduction_options.copy()
+
+        for channel in channels:
+            if channel is None:
+                continue
+            elif isinstance(channel, self.reflect_klass):
+                pass
+            else:
+                try:
+                    channel = os.path.join(data_folder, channel)
+                except TypeError:
+                    # original channel is not a string
+                    pass
+                finally:
+                    # let's hope it's an h5 file
+                    channel = self.reflect_klass(channel)
+
+            if channel.spin_state is SpinChannel.DOWN_DOWN:
+                self.dd = channel
+            elif channel.spin_state is SpinChannel.DOWN_UP:
+                self.du = channel
+            elif channel.spin_state is SpinChannel.UP_DOWN:
+                self.ud = channel
+            elif channel.spin_state is SpinChannel.UP_UP:
+                self.uu = channel
+
+        assert (
+            self.dd is not None
+        ), "down_down spin channel is not SpinChannel.DOWN_DOWN!"
+        assert (
+            self.uu is not None
+        ), "up_up spin channel is not SpinChannel.UP_UP!"
+
+    @property
+    def spin_channels(self):
+        return [
+            s.spin_state.value if s is not None else None
+            for s in [self.dd, self.du, self.ud, self.uu]
+        ]
+
+    def process_beams(self, reduction_options=None):
+        """
+        Process beams in SpinSet.
+
+        Reduction options for each spin channel are specified by
+        self.dd_opts, self.du_opts, self.ud_opts, and self.uu_opts where
+        a standard set of options is provided when constructing the object.
+        To specify different options for each spin channel (such as using
+        the ManualBeamFinder for only spin-flip channels), update the
+        reduction options for the specific spin channel in SpinSet, then
+        process the beams. i.e.
+
+        from refnx.reduce.manual_beam_finder import ManualBeamFinder
+        mbf = ManualBeamFinder()
+
+        spinset = SpinSet(
+            "PLP0051296.nx.hdf",
+            "PLP0051294.nx.hdf",
+            up_down="PLP0051295.nx.hdf",
+            down_up="PLP0051297.nx.hdf",
+            data_folder=data_dir
+        )
+        spinset.du_opts.update({"manual_beam_find" : mbf, peak_pos : -1})
+        spinset.process_beams()
+
+        Parameters
+        ----------
+        reduction_options : dict
+            A single dict of options used to process all spectra. If
+            this is None, then process_beams will use individual dicts
+            for each spin channel
+        """
+
+        if reduction_options:
+            print(
+                "Applying the supplied reduction_options to all spin channels"
+            )
+            self.dd_opts = reduction_options.copy()
+            self.du_opts = reduction_options.copy()
+            self.ud_opts = reduction_options.copy()
+            self.uu_opts = reduction_options.copy()
+
+        for opts, beam in zip(
+            [self.dd_opts, self.du_opts, self.ud_opts, self.uu_opts],
+            [self.dd, self.du, self.ud, self.uu],
+        ):
+            if beam is None:
+                continue
+            else:
+                beam.process(**opts)
+
+    def plot_spectra(self, **kwargs):
+        """
+        Plots the processed spectrums for each spin state in the SpinSet
+
+        Requires matplotlib to be installed
+        """
+        import matplotlib.pyplot as plt
+
+        fig, ax = plt.subplots()
+        ax.set(xlabel="Wavelength ($\\AA$)", ylabel="Intensity (a.u.)")
+
+        for spinch in [self.dd, self.du, self.ud, self.uu]:
+            if spinch is None:
+                continue
+            x = spinch.processed_spectrum["m_lambda"][0]
+            y = spinch.processed_spectrum["m_spec"][0]
+            yerr = spinch.processed_spectrum["m_spec_sd"][0]
+            ax.errorbar(x, y, yerr, label=spinch.cat.sample_name)
+        return fig, ax
 
 
 def basename_datafile(pth):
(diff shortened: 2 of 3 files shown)


refnx/reduce/test/test_platypusnexus.py
```diff
@@ -374,6 +375,12 @@ def test_PNR_metadata(self):
         assert_almost_equal(self.f8864.cat.cat["anal_flip_current"], 0)
         assert_almost_equal(self.f8864.cat.cat["pol_flip_current"], 0)
 
+        # Check SpinChannel for each file
+        assert_(self.f8861.spin_state == SpinChannel.UPUP)
```
andyfaff, line 379: make them into plain asserts: `assert self.f8861.spin_state == SpinChannel.UPUP`

refnx/reduce/__init__.py
```diff
@@ -14,6 +14,9 @@
     Catalogue,
     SpatzNexus,
     ReductionOptions,
+    SpinChannel,
+    SpinSet,
+    _not_none,
```
andyfaff, line 19: don't add `_not_none` to `__init__.py`, probably don't want it exported.

refnx/reduce/platypusnexus.py
```diff
@@ -626,6 +627,118 @@ def _check_sample_environments(self, d, h5d):
         return d
 
 
+class SpinChannel(Enum):
+    """
+    Describes the spin state of a polarised neutron beam.
+    """
+
+    UPUP = (1, 1)
```
andyfaff, line 635: In places we have `PLUS/MINUS`, and elsewhere `UP/DOWN`. We should standardise throughout refnx on one nomenclature.

refnx/reduce/platypusnexus.py
```diff
@@ -626,6 +627,118 @@ def _check_sample_environments(self, d, h5d):
...
+    UPDOWN = (1, 0)
```
andyfaff, line 636: is `UP_DOWN` preferable to `UPDOWN`? My preference would be to use the former throughout.

refnx/reduce/platypusnexus.py
```diff
@@ -626,6 +627,118 @@ def _check_sample_environments(self, d, h5d):
...
+    ----------
+    list of {str, h5data}
+        list of str, or list of h5py file handles pointing to
+        a set of polarised neutron beam files
+
+    data_folder: {str, Path}
+        Path to the data folder containing the data to be reduced.
+    """
+
+    def __init__(self, dataset, data_folder=None):
```
andyfaff, line 657: Creating a `SpinSet` requires mandatory UPUP and DOWNDOWN. I think it would be cleaner to have the function signature as:

```
# or down_down, up_up, ...
def __init__(self, up_up, down_down, up_down=None, down_up=None, data_folder=None):
```
because then the SpinSet cannot be created without providing at least two of the spin states.

refnx/reduce/platypusnexus.py
```diff
@@ -626,6 +627,118 @@ def _check_sample_environments(self, d, h5d):
...
+
+        # Currently only Platypus has polarisation elements
+        self.reflect_klass = PlatypusNexus
+
+        self.data_folder = data_folder
+        if len(dataset) > 4:
```
andyfaff, line 663: If we change the function signature this won't be required, but it would be important to check that up_up and down_down, etc, represented the correct spin states.

refnx/reduce/platypusnexus.py
```diff
@@ -626,6 +627,118 @@ def _check_sample_environments(self, d, h5d):
...
+        if isinstance(dataset[0], PlatypusNexus):
+            self.beams = self.arrange_spin_channels(dataset)
+        elif type(dataset[0]) is str:
+            fnames = [os.path.join(self.data_folder, dset) for dset in dataset]
+            PLPset = [self.reflect_klass(d) for d in fnames]
+
+            self.beams = self.arrange_spin_channels(*PLPset)
+        else:
+            raise ValueError("Dataset should be list of PlatypusNexus or str")
+        self.mm, self.mp, self.pm, self.pp = self.beams
```
andyfaff, line 678: Settle on nomenclature - are we using down/up or minus/plus?

refnx/reduce/platypusnexus.py
```diff
@@ -626,6 +627,118 @@ def _check_sample_environments(self, d, h5d):
...
+        self.data_folder = data_folder
+        if len(dataset) > 4:
+            raise ValueError("Too many spin channels!")
+        if len(dataset) < 2:
+            raise ValueError(
+                "Too few spin channels! Need at least R++ and R--."
+            )
+        if isinstance(dataset[0], PlatypusNexus):
+            self.beams = self.arrange_spin_channels(dataset)
+        elif type(dataset[0]) is str:
```
andyfaff, line 671:
````
beams = []
for spinch in [down_down, up_up, down_up, up_down]:
    if spinch is None:
        beams.append(None)
    elif isinstance(spinch, PlatypusNexus):
        beams.append(spinch)
    else:
        pn = self.reflect_klass(spinch)
        beams.append(pn)
    
beams = self._arrange_spin_channels(beams)
self.mm, self.mp, self.pm, self.pp = *beams
# make a check that mm/pp/mp/pm are the spin states they say there are, i.e. the user input to the constructor is correct.
```

andyfaff: probably not a need to make `beams` an attribute.

refnx/reduce/platypusnexus.py
```diff
@@ -626,6 +627,118 @@ def _check_sample_environments(self, d, h5d):
...
+        self.mm, self.mp, self.pm, self.pp = self.beams
+
+    @property
+    def spin_channels(self):
+        return [
+            s.spin_state.value if s is not None else None
+            for s in [self.mm, self.mp, self.pm, self.pp]
+        ]
+
+    def arrange_spin_channels(self, *data):
```
andyfaff, line 687: `def _arrange_spin_channels(self, beams):`

refnx/reduce/__init__.py
andyfaff: actually, function is not required at all.

refnx/reduce/platypusnexus.py
```diff
@@ -626,6 +627,118 @@ def _check_sample_environments(self, d, h5d):
...
+        """
+        Function that takes a random input of spin channels at
+        a single angle and returns them arranged according to
+        --, -+, +-, ++ and replaces the spin channel with None
+        if it was not measured.
+        """
+        states = [None] * 4
+
+        for a in _not_none(*data):
```
andyfaff, line 696:
```
for i, a in enumerate(data):
    if a is None:
        continue
    elif a.spin_state is SpinChannel.DOWN_DOWN:
        states[0] = a
    ...

assert states[0] is not None, "A down/down spin channel was not provided"
assert states[3] is not None, "An up/up spin channel was not provided"

return states
```

refnx/reduce/platypusnexus.py
```diff
@@ -626,6 +627,118 @@ def _check_sample_environments(self, d, h5d):
...
+        fig, ax = self.mm.plot(**kwargs)
+
+        for spinch in _not_none(self.mp, self.pm, self.pp):
+            fig, ax = spinch.plot(fig=fig, **kwargs)
+
+        fig.legend()
+        return fig, ax
+
+
+def _not_none(*arrays):
```
andyfaff, line 735: remove this function.

refnx/reduce/platypusnexus.py
```diff
@@ -1909,9 +2022,20 @@ def __init__(self, h5data):
         self.prefix = "PLP"
         with _possibly_open_hdf_file(h5data, "r") as f:
             self.cat = PlatypusCatalogue(f)
-            if self.cat.mode == "POLANAL":
+            if self.cat.mode in ["POL", "POLANAL"]:
                 self.cat = PolarisedCatalogue(f)
 
+                # Set spin channels based of flipper statuses
+                if self.cat.pol_flip_current:
```
andyfaff, line 2029: can we flatten this out?

```
if self.cat.pol_flip_current and self.cat.anal_flip_current:
     self.spin_state = SpinChannel.UP_UP
elif self.cat.pol_flip_current and not self.cat.anal_flip_current:
    self.spin_state = SpinChannel.UP_DOWN
...
```

refnx/reduce/platypusnexus.py
```diff
@@ -626,6 +627,118 @@ def _check_sample_environments(self, d, h5d):
...
+    def process_beams(self, reduction_options=None):
+
+        if reduction_options is None:
+            reduction_options = {
+                "lo_wavelength": 2.5,
+                "hi_wavelength": 12.5,
+                "rebin_percent": 3,
+            }
+
+        for beam in _not_none(self.mm, self.mp, self.pm, self.pp):
```
andyfaff, line 717: Hmm, what happens if you want to use different options for each state? The wavelength/rebin_percent should be the same, but people might want to specify peak positions, etc, which would require a different reduction options for each spin state.

andyfaff (changes requested):

refnx/reduce/platypusnexus.py
ohcpaull: I figure in this case, the user can process each spin state manually using something like `spinset.dd.process(**reduction_options_dd)`. I think it makes more sense for `process_beams` to assume identical options for processing rather than having to worry about the order of a list of `ReductionOptions` passed to it

andyfaff: The different beams have to be processed onto the same wavelength axis, which is the benefit of using identical options. However, I was wondering about what happens during location of the specular peak. For the NSF channels the auto-peak location may work really well, but for the SF channels it may prove troublesome.

Thus, one may want to use a mix of automated peak finding and manual beam finding for different spin channels.

andyfaff:
Something is going on with the commits in git here. Don't know what exactly. Can you squash down into 1?

andyfaff:
There's also no test for `SpinSet` yet.

refnx/reduce/platypusnexus.py
ohcpaull: What about adding `manual_beam_find : ManualBeamFinder()` and `peak_pos : -1` to `reduction_options` if a SF channel is present?

andyfaff:
I've made some changes to the PR, with a little bit of rationalisation. If you could add tests for the SpinSet that would be awesome.

refnx/reduce/platypusnexus.py
ohcpaull: I seem to be having trouble using ManualBeamFinder lately, it makes Python stop working...

andyfaff: How are you using it? In Jupyter:

```
%gui qt
# IMPORTANT: wait for a few seconds

from refnx.reduce.manual_beam_finder import ManualBeamFinder

mbf = ManualBeamFinder()
```

ohcpaull: I am sure I had that (including %gui qt) the other day and was still having problems, but seems to be working now. All good.

ohcpaull:
> 
> 
> Something is going on with the commits in git here. Don't know what exactly. Can you squash down into 1?

Thanks for spotting this - I think there was an accidental merge with a different branch

refnx/reduce/platypusnexus.py
ohcpaull: Hi Andy, I have now added reduction options as attributes in SpinSet for each spin channel in one of the latest commits. There are some failed tests but from my understanding they don't seem to be because of my code (maybe an outdated package dependency?).

ohcpaull: Hi Andy, I have now added reduction options as attributes in the SpinSet class for each spin channel. The user can create the SpinSet object, then update the individual reduction options as they please, then process the spectra with those options. One or two tests have failed but I think this is due to a package dependency or something

refnx/reduce/platypusnexus.py
```diff
@@ -626,6 +627,166 @@ def _check_sample_environments(self, d, h5d):
...
+        ), "up_up spin channel is not SpinChannel.UP_UP!"
+
+    @property
+    def spin_channels(self):
+        return [
+            s.spin_state.value if s is not None else None
+            for s in [self.dd, self.du, self.ud, self.uu]
+        ]
+
+    def process_beams(self, reduction_options=None):
```
andyfaff, line 718: `process`

refnx/reduce/platypusnexus.py
```diff
@@ -626,6 +627,166 @@ def _check_sample_environments(self, d, h5d):
...
+        """
+        Process beams in SpinSet.
+
+        Reduction options for each spin channel are specified by
+        self.dd_opts, self.du_opts, self.ud_opts, and self.uu_opts where
```
andyfaff, line 723: `SpinSet.dd_opts`

refnx/reduce/platypusnexus.py
```diff
@@ -626,6 +627,166 @@ def _check_sample_environments(self, d, h5d):
...
+            )
+            self.dd_opts = reduction_options.copy()
+            self.du_opts = reduction_options.copy()
+            self.ud_opts = reduction_options.copy()
+            self.uu_opts = reduction_options.copy()
+
+        for opts, beam in zip(
+            [self.dd_opts, self.du_opts, self.ud_opts, self.uu_opts],
+            [self.dd, self.du, self.ud, self.uu],
+        ):
```
andyfaff, line 763: check for options that must be the same. It'll be required that the wavelength bins are the same.

refnx/reduce/platypusnexus.py
```diff
@@ -626,6 +627,166 @@ def _check_sample_environments(self, d, h5d):
...
+    UP_DOWN = (1, 0)
+    DOWN_UP = (0, 1)
+    DOWN_DOWN = (0, 0)
+
+
+class SpinSet(object):
+    """
+    Describes a set of spin-channels at a given angle.
+
+    TODO: implement polarisation efficiency correction within this class
```
andyfaff, line 645: No documentation is given to each of the reduction option attributes that are possible.

andyfaff: i.e. `du_opts`

refnx/reduce/platypusnexus.py
andyfaff: and add to documentation which attributes must be the same.

andyfaff:
@ohcpaull, I made some changes to this PR and force pushed to your branch, followed by a squash. Unfortunately I made a mistake on the squash and it lost the correct authorship attribution. I managed to edit the author, but whilst it looks ok on my local copy of the repo it doesn't come up on GitHub properly.

The only real way of restoring that is for me to roll back the commits that have just been added in this PR. Then for you to force push your local copy of the `spinset_classes` branch back to GH, then I have another go at merging them.

What would you like to do?

PR #583, opened 2021-07-29
Remove `data_folder` kwd from `SpinSet` initialisation

andyfaff:
SpinSets should be constructed with a fully qualified filename. This removes the data_folder keyword argument.


andyfaff:
This means that loading files from cwd should work.
@ohcpaull @dcortie

PR #613, opened 2022-01-17
MAINT: add more keys to reduction output dict

andyfaff:


andyfaff:
@llimeht this doesn't fix the fact that `reflected_beam` state is present in the `PlatypusReduce` file. However, it does ameliorate things a little by placing that information in the reduction dict that is output from the `reduce` method.

PR #620, opened 2022-02-22
A little energy dispersive machinery

andyfaff:
@igresh @haydenrob a little energy dispersive machinery. Existing refnx code should 'just work'. However, refellips is going to have to be adjusted to work with the changes here.

`Component` that use Scatterers have to be rewritten slightly, to use `Scatterer.complex(getattr(structure, "wavelength", None))` when trying to calculate complex SLD/RI.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -400,5 +400,15 @@ in 0.1.13.
 - added fields to ReflectReduce output dict, including omega, reflected_beam,
   direct_beam.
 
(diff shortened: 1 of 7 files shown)


andyfaff:
Ideally the energy dispersive-ness would be vectorised over for speed, but this should do the job.

haydenrob:
That's great - thanks Andy. I'll check it out this afternoon.

PR #643, opened 2022-09-07
changes required for PyQt6

andyfaff:


andyfaff:
Perhaps only consider when PyQt=6 is available on conda-forge

andyfaff:
Change to qtpy which should allow PySide6

PR #671, opened 2023-07-02
New blend_data function to combine weighted datasets, and small method changes for Data1D

mattgebert:
Added a new function to blend data, which statistically combines overlapping data using y-err weighting. This is unique functionality and different to the truncate options of add_data, allowing seamless combination of datasets where y-err is optimized.

Also added copy functions to data1D & parent classes. Copy method is used to isolate raw data arrays. Metadata, Orso, sld_profile and datafilenumber are shallow copied.

Lastly, ensured that the initializer doesn't always set `self._mask` to `None`, especially if copying a dataset (passed through the constructor).

### DIFF
--- a/refnx/dataset/data1d.py
+++ b/refnx/dataset/data1d.py
@@ -395,12 +395,15 @@ class Data1D:
         Parameters
         ----------
         other : Data1D
-            Additional dataset. Point spacing in Q should match existing dataset, so overlapping points can be combined.
+            Additional dataset. Point spacing in Q can be different.
         
         Notes
         -----
-        self.x is interpolated to match other.x in overlap reigon, before blending.
-        Raises `AttributeError` if different data density in overlap reigon.
+        Overlap points are calculated for each dataset.
+        The shorter overlap dataset is interpolated to match the larger overlap dataset.
+        Then a weighted average of dataset values are calculated, favoring smaller y-error values.
+        Errors are also propogated throughout calculations.
+        
         Raises `AttributeError` if neither dataset has a y_err attribute.
         Raises `ValueError` if no overlap in x; should use add_data.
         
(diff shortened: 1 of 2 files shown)


andyfaff:
I'm on leave this week, so won't get around to this for a week. With regard to working out a scale factor is there any functionality that https://github.com/refnx/refnx/blob/main/refnx/util/nsplice.py doesn't cover?

mattgebert:
> With regard to working out a scale factor is there any functionality that https://github.com/refnx/refnx/blob/main/refnx/util/nsplice.py doesn't cover?
Nope! In fact I used your function `get_scaling_in_overlap` to calculate the scaling before blending the overlap domain. 😄

andyfaff:
First of all, thank you for contributing to `refnx`. All contribs are welcomed. Unfortunately for this PR it's unclear what functionality is introduced over and above what is already present in `refnx`.

It seems that the core of the PR is adding the `blend_data` function. I believe this functionality is already present in the `add_data` method. If scaling is required when splicing simply set `requires_splice=True`. This scaling takes into account error bars. Perhaps the keyword would've been better named as `requires_scaling`, but that horse has bolted.
In addition, if you don't want to remove the high Q end of the first dataset in the overlap region (the truncating), you can set `trim_trailing=False`.

If there is other functionality being introduced in the PR we should do it in separate non-overlapping PRs, it makes the review workflow easier. Lastly, it's probably a good idea to create an issue first before creating a PR. It saves the work of writing code for a PR that may need a lot of modification before merging, and it allows me to give advice on whether the issue is solvable in another way. An alternative to an issue may simply be a video call which has a higher bandwidth.

mattgebert:
> If there is other functionality being introduced in the PR we should do it in separate non-overlapping PRs, it makes the review workflow easier. Lastly, it's probably a good idea to create an issue first before creating a PR. It saves the work of writing code for a PR that may need a lot of modification before merging, and it allows me to give advice on whether the issue is solvable in another way. 

No problem! I haven't contributed much to online repos, apologies for my lack of experience. It might have made more sense to contact you in advance, or as you suggest raise an issue. I was too eager to contribute 🥲 

> I believe this functionality is already present in the `add_data` method. 

Personally I don't believe this is the case. `Add_data` offers the following possibilities:

1. `requires_splice` option: When the new data is added to the dataset do you want to scale it vertically so that it overlaps with the existing data? y and y_err in data_tuple are both multiplied by the scaling factor.
2. `trim_trailing` option: When the new data is concatenated do you want to remove points from the existing data that are in the overlap region? This might be done because the datapoints in the data_tuple you are adding have have lower y_err than the preceding data.

The issue with the trim trailing option is the lack of optimized data addition between various datasets. Take for example the following data.
01 Blend-Example1-Y
Lets scale it appropriately...
02 Blend-Example1-Scaled
03 Blend-Example1-Yerrs
When scaled, the y-errs are not globally superior to the first dataset in the overlap region. This means that either option of True/False for `trim trailing` will lead to an unoptimized result. 

Instead we can statistically merge the overlap regions, weighted by y-errs, leading to lower probabilities and a more normalized result within reduction bin windows.
04 Blend-Example1-vs-AddData

Hopefully this example demonstrates the additional functionality, smoothly joining between datasets.

Let me know if you still don't think this is worthwhile, or would rather place the functionality elsewhere (or even incorporate it as a different option to add_data).

andyfaff:
If I understand correctly you're averaging (with weighting) points from the 1st and 2nd angles that are at the same Q-value?  There are a few points to mention here:

1. Untrimmed+merged data would be the same statistically as your solution, except it looks slightly messier. It's the fault of a reviewer if they think the data looks noisy.
2. The points from the first and second set are unlikely to coincide at the same Q-point. In order to get points coinciding I think you're using interpolation, which I'm not a fan of (slight massage, even if it can be ultimately justified). There's an additional issue, the Q-resolution for the two points may not be equivalent. One data point may have narrow resolution, one point may have wide resolution; so one has to figure out what the correct Q-resolution would be. Even though the fractional Q-resolution should be constant there are also modes of Platypus+Spatz where it's not.
3.  Possibly a better way to handle this is to use a rebinning procedure. Currently the rebinning step is done earlier, on the TOF spectrum. Let's say we don't rebin at that point, but we rebin the reflectivity curves instead. This would also produce something smoother. However, this also means that the correct resolution function has to be figured out (any rebinning coarsens Q-resolution). Furthermore, background subtraction (which is done on a wavelength by wavelength basis of a tof-rebinned detector image) becomes noisier.
4. Perhaps the best option is to use the current reduction procedure, then do an additional rebinning step on the red curve you display in your last post.

mattgebert:
I see - thanks for your feedback, and patience, Andy. 

1.  Agreed.
2. The algorithm does interpolate to match Q values, before doing a weighted average over the same Q values. However, x-uncertainty (Q-resolution) is not averaged, it's calculated through proper error propagation of the averaging function.
3. Haha yeah this makes sense - I would love to dig into this but this will probably be a bit too time consuming for me at the moment 🥲 as I'm writing up my thesis. 
4. Good suggestion.

I shall close this pull request soon, as it is not necessary.

PR #672, opened 2023-07-19
DOC: pip install refnx[all]

andyfaff:


andyfaff:
@dcortie points out that `pip install refnx[all]` may be a simpler way of installing all optional dependencies.

andyfaff:
Had to make some extra changes caused by cython3 being released.

PR #695, opened 2023-09-07
TST: update conftest

andyfaff:


refnx/conftest.py
```diff
@@ -22,21 +22,21 @@ def data_directory(tmpdir_factory):
         returned. If the retrieval fails then None is returned.
     """
     url = "https://github.com/refnx/refnx-testdata/archive/master.zip"
-    tmpdir = tmpdir_factory.mktemp("data")
+    tmpdir = Path(tmpdir_factory.mktemp("data"))
```
llimeht, line 25: Switching to pytest's `tmp_path_factory` might be even easier: https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html#the-tmp-path-factory-fixture

refnx/reduce/test/test_batch_reduce.py
```diff
@@ -19,19 +19,19 @@ class TestReduce:
     @pytest.mark.usefixtures("no_data_directory")
     @pytest.fixture(autouse=True)
     def setup_method(self, tmpdir, data_directory):
-        self.pth = pjoin(data_directory, "reduce")
+        self.pth = Path(data_directory) / "reduce"
```
llimeht, line 22: This should already be a `Path` from the fixture, shouldn't it?

refnx/reduce/test/test_batch_reduce.py
```diff
@@ -19,19 +19,19 @@ class TestReduce:
...
         self.cwd = os.getcwd()
 
-        self.tmpdir = tmpdir.strpath
+        self.tmpdir = Path(tmpdir.strpath)
```
llimeht, line 25: Maybe switch tmpdir to `tmp_path` while at it? https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html#the-tmp-path-fixture
(also no need to go via `strpath` then

refnx/reduce/test/test_batch_reduce.py
```diff
@@ -169,25 +168,25 @@ def test_persistence(self, tmpdir):
 
         # presence of this file would indicate a failure of
         # the previous persistent=False
-        assert not os.path.exists(self.cache._default_persistent_cache)
+        assert not Path(self.cache._default_persistent_cache).exists()
```
llimeht, line 171: Probably should rewrite all the batch reducer in terms of pathlib too (some day)

refnx/reflect/_app/test/test_app.py
```diff
@@ -83,16 +82,16 @@ def test_add_spline_save(qtbot, tmpdir):
...
 def save_and_reload_experiment(app, tmpdir):
     # save and reopen experiment.
-    sf = pjoin(str(tmpdir), "experiment1.mtft")
+    sf = Path(tmpdir) / "experiment1.mtft"
     # this is just to make sure that the file exists
     with open(sf, "wb") as f:
         f.write(b"sksij")
 
-    app.settings.experiment_file_name = sf
+    app.settings.experiment_file_name = str(sf)
```
llimeht, line 90: Is it worth broadening that property to accept Path objects directly?

andyfaff: Not really, but I've just made it insensitive.

refnx/dataset/test/test_reflectdataset.py
```diff
@@ -115,18 +119,26 @@ def test_loading_junk(self):
         from pytest import raises
 
         with raises(RuntimeError):
-            ReflectDataset(os.path.join(self.pth, "../__init__.py"))
+            ReflectDataset(self.pth / "../__init__.py")
```
llimeht, line 122: perhaps `ReflectDataset(self.pth.parent / "__init__.py")` (saves potential portability issues with assuming dirsep)

refnx/dataset/test/test_reflectdataset.py
```diff
@@ -261,25 +273,25 @@ def test_mask(self):
...
 class TestData1D:
     @pytest.fixture(autouse=True)
-    def setup_method(self, tmpdir):
-        self.pth = os.path.dirname(os.path.abspath(__file__))
-        self.cwd = os.getcwd()
-        self.tmpdir = tmpdir.strpath
-        os.chdir(self.tmpdir)
+    def setup_method(self, tmp_path):
+        self.pth = Path(__file__).absolute().parent
+        self.cwd = Path(".").resolve()
```
llimeht, line 292: `Path.cwd()`; no idea how portable `.` is as cwd.  (There are lots of other places where `Path(".")` was used; I wasn't sure if some of them should be based on `__file__` rather than cwd)

A problem for another day: this doesn't actually look like a pytest [fixture](https://docs.pytest.org/en/latest/how-to/fixtures.html), rather compatibility layer to nose/xunit. Combining the setup/teardown methods into one, yielding in the middle, and adding that function to the test methods that need it would be the pytest fixture idiom.

refnx/dataset/test/test_dataset.py
```diff
@@ -1,7 +1,8 @@
-import os.path
 from pathlib import Path
-import pytest
 import glob
```
llimeht, line 2: In `test_load_data` (below), pathlib's glob can be used id you want, then omitting this import. https://docs.python.org/3/library/pathlib.html#pathlib.Path.glob

refnx/reduce/_app/__init__.py
```diff
@@ -10,7 +10,7 @@ def gui():
 
     time_str = time.strftime("%Y%m%d-%H%M%S")
     log_filename = "slim_" + time_str + ".log"
-    log_filename = os.path.join(os.path.expanduser("~"), log_filename)
+    log_filename = Path("~").expanduser() / log_filename
```
llimeht, line 13: `Path.home()` https://docs.python.org/3/library/pathlib.html#pathlib.Path.home

llimeht (commented):
My, you have had fun :smile:

refnx/dataset/test/test_reflectdataset.py
andyfaff: I assume `tmp_path` is a fixture, https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html#the-tmp-path-fixture

llimeht: Yes it is, but the `pytest.fixture` decorator is for when you create new fixtures, not for when you use fixtures.

There's no need for a decorator there to cause the `tmp_path` fixture to be used, that's done behind the scenes by pytest itself. pytest sees `tmp_path` in the method signature and invokes the fixture for it, feeding in the result.

If a fixture is specifically requested by including it in the signature of the methods that use it, then that fixture just needs `@pytest.fixture`.  If you want a fixture to be run all the time without including it in the signatures of each method, then the `autouse=True` parameter arranges that within that scope. 

In this case, decorating `setup_method` as an autouse fixture makes it like a fixture for all test methods within the class, but it's not really written like a fixture - it's written like an old-school pair of setup/teardown methods. If that decoration were actually needed, then I suspect the cleanup code is never run (because there's no end-of-fixture hook like that; that's what a yield fixture is for). 

Rather than this actually being a fixture, this could be run with one of pytest's other compatibility shims that exist to help convert test suites from other testing approaches. The setup_method/teardown_method idiom is one of them, and fixture marking isn't needed for that. Similarly, pytest also has `unittest` compatibility that lets you inherit from `unitest.TestCase` and have `setUp` methods (but fixtures don't work with that very well). 

Probably, however, you want this to be a `chdir_tmp_path` fixture that requests the `tmp_path` fixture, chdirs to it, yields, then chdirs back. Some error handling on that is easier, `self` is actually unneeded for that too.

refnx/dataset/test/test_reflectdataset.py
```diff
@@ -115,18 +119,26 @@ def test_loading_junk(self):
         from pytest import raises
 
         with raises(RuntimeError):
-            ReflectDataset(os.path.join(self.pth, "../__init__.py"))
+            ReflectDataset(self.pth / "../__init__.py")
+
+    def test_bytes(self):
+        with open(self.pth / "c_PLP0033831.txt", "rb") as f:
+            data = f.read()
+        s = data.decode().replace("\r", "\n")
```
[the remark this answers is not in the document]
andyfaff: I'm not worried about reading Bytes, I was wanting to make sure that a StringIO would work.

refnx/dataset/test/test_reflectdataset.py
andyfaff: when I get rid of all the decorators nothing works. I don't understand what the correct thing is to do here.

llimeht: `setup_method` currently needs to be annotated because it doesn't have the method signature signature that is required for the xunit compat shim to kick in. `tear_down` does use that compat shim because it has the right signature. That works, but is unusual, and one has to remember why it works for when it breaks one day :)

For TestData1D, the following is more idiomatic:
```python
    @pytest.fixture(autouse=True)
    def tmp_path_chdir(self, tmp_path):
        cwd = Path.cwd()
        os.chdir(tmp_path)

        try:
            yield
        finally:
            os.chdir(cwd)
```

It is possible to go one step further than that and make it an explicit use fixture rather than an autouse fixture; that only involves adding `tmp_path_chdir` to the function definitions of each of the functions that need it. Given how often this particular sequence is used (11 times in the test suite, it seems), it might be worth making it an explicitly used fixture and adding it to `conftest.py` so that it's not repeated multiple times.

Converting the `setup_method` in `TestReflectDataset` could be done by splitting it into three - each fixture only does one thing, and no side effects:
```python
    @pytest.fixture(autouse=True)
    def tmp_path_chdir(self, tmp_path):
        cwd = Path.cwd()
        os.chdir(tmp_path)

        try:
            yield
        finally:
            os.chdir(cwd)

    @pytest.fixture
    def dummy_data(self):
        data = ReflectDataset()

        x1 = np.linspace(0, 10, 115)
        y1 = 2 * x1
        e1 = np.ones_like(x1)
        dx1 = np.ones_like(x1)
        data.add_data((x1, y1, e1, dx1))
        yield data

    @pytest.fixture
    def data_location(self):
        location = Path(__file__).absolute().parent
        yield location

    def test_load(self, data_location):
        # load dataset from XML, via file handle
        dataset = ReflectDataset()
        with open(data_location / "c_PLP0000708.xml") as f:
            dataset.load(f)
```

`unittest` is very focused on classes providing scope for setup/teardown methods, it encourages the use of `self` throughout the tests to store parameters and data related to the harness. The setup/teardown interaction with the test functions relies on side-effects by storing state in `self`, but that also reduces isolation between test functions and so tests can accidentally influence each other. `pytest` instead manages it all via fixtures and the use of `self` is normally not needed or wanted. With `pytest`, classes are (more or less) only used for an additional level of scoping to permit easy selection of sets of tests on the commandline, for applying marks, and sometimes for autouse fixtures

(I have this particular file all updated to this more idiomatic fixture use if you want a PR for it; there's no need to update all at once, it can be a "when next touching these tests" process.)

PR #787, opened 2024-11-14
ENH: appropriate unit conversion when loading ORSO data

andyfaff:
@maxskoda, here is the unit conversion '1/nm' --> '1/angstrom' you were asking for.  The default way of loading a dataset should probably be via the `refnx.dataset.load_data` function.


andyfaff:
Specifically this line: https://github.com/refnx/refnx/pull/787/files#diff-0d466de21f32c513668c588a57ac76381aad2d23c314ce058df1dae4d7763550R213

maxskoda:
Many thanks @andyfaff - as always you are very fast indeed!

PR #825, opened 2025-03-18
MAINT: fix spinset raise warning

DLCortie:
line 779, add raise so warning is output correctly


andyfaff:
gr8, can you add a test?

PR #888, opened 2026-04-20
DOC: non-gaussian resolution function example.

andyfaff:


andyfaff:
@jfkcooper, this example might be useful for the ESS reflectometers at some point.

jfkcooper:
Thank you - yes!

andyfaff:
Are there other specific features that might be required for ESS data analysis?

jfkcooper:
One thing which comes to mind is functional forms of backgrounds (if this isn't already a thing). We will be integrating over a large area of the detector, and a flat background in theta will not be in Q, so we may run into some 'fun', and my preference is to model backgrounds rather than subtract them. Is this currently possible?

andyfaff:
I think that it would be possible to add a functional background.

andyfaff:
What number of parameters, and what kind of function? Something like `numpy.polynomial.Polynomial`?

jfkcooper:
That seems like it would cover it well. I was thinking of something simple, but being a little more flexible and pushing off the calculation to numpy seems a better way forward

PR #938, opened 2026-08-05
MAINT: win exe size

andyfaff:


andyfaff:
The windows exe size has been reduced to a more manageable 120 Mb, down from 390 Mb!

PR #949, opened 2026-09-16
Add SPDX annotations to source files

llimeht:
As discussed elsewhere, this PR

- fixes up the refnx licensing so that it is all under the same licence
- adds SPDX headers to the source code that state licence and attribution

I've not added SPDX headers to the vendored code in `_lib`.

With this in place, various licence-reconciling software generates many fewer false positives about copyright holders and licences.


andyfaff:
How important is the start-end year statement?

There are a few clang-format lint things need fixing.

llimeht:
> How important is the start-end year statement?

Depends who you ask.

It's common to set them to be the dates for the individual file. It's common to set them to the dates for the project (knowing that you'll have copied or move stuff from file to file within the project). It's common to omit them entirely inside the files and point only at the LICENSE file (which I note has not been updated for a while and whose dates don't match the dates in git history). There are strongly held opinions on each option. In practice, the dates are irrelevant with how copyright is extended and managed.

andyfaff:
thank you @llimeht

PR #950, opened 2026-09-16
Lint fixes in examples

llimeht:
Three simple fixes for the examples

- `examples/mpi_parallelisation.py` is a Python script but currently has `#!/bin/sh` 
- two example scripts had no `#!` at all
- three example scripts weren't marked as executable

These haven't caused issues before because they can be run as `python foo.py`, but they are executable directly too.

(Flagged by Debian's package linter, `lintian`)


andyfaff:
These scripts are v. old! I still remember writing them.

